PR #3, opened 2017-12-01
i600 - use vaultr's new caching support

richfitz:
This should put the work in https://github.com/vimc/vaultr/pull/2 a bit more into context


richfitz:
Done in https://github.com/vimc/montagu-reports/pull/71 and by putting the orderly docs online: https://vimc.github.io/orderly/articles/orderly.html#customising-the-configuration

PR #7, opened 2018-05-15
i1754 - Correct documentation

richfitz:
This is used by orderly::orderly_new and is incorrect


richfitz:
I have no idea why this is failing on travis, but I do not think that it is to do with this PR.  I will investigate further...

richfitz:
This is ready for a look @JamesThompson1729

PR #10, opened 2018-05-18
I1756 - option to find reports from a filesystem

richfitz:
Requires https://github.com/vimc/montagu-r/pull/4

This is needed for the ebola work


richfitz:
This will fail until vimc/montagu-r#4 is merged

PR #12, opened 2018-05-22
I1772 - support basic auth and non-montagu servers

richfitz:
This work needs https://github.com/vimc/montagu-r/pull/5

This is primarily to support the ebola response and extends the orderly remote functions a bit to support non-VIMC servers.  It's going to need a further round of rationalisation once we really disentangle montagu and orderly.

The `orderly_config.yml` picks up additional options

- `vault_server` - the address of the vault that we should use
- `api_server` - a set of remote api servers

Most of the logic is within the validation function in config.R and also in `orderly_remote_api_server`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ after_success:
   - Rscript -e 'covr::codecov()'
 r_github_packages:
   - ropensci/cyphr
-  - vimc/montagu-r@i1772
+  - vimc/montagu-r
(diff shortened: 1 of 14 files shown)


richfitz:
appveyor is failing because of the montagu-r dependency.  https://github.com/vimc/montagu-r/pull/5 is merged ping me and I'll update the travis config too, but the build there will start working at that point

PR #18, opened 2018-06-04
i1800: Update to use R 3.5.0

richfitz:
There will be a chain of updates here as this flows through other containers.  This is the base one.

When the docker containers are updated you need to check teamcity to make sure they're building ok


richfitz:
The final step of this one is https://github.com/vimc/orderly.server/pull/6 and then updating montagu itself

PR #21, opened 2018-06-04
i1796: don't download nonexistant reports

richfitz:
We discussed this a while ago after being bitten by it, and here is a fix!

Try and download nonexistant reports before and after applying this PR


R/remote_api.R
```diff
@@ -6,29 +6,41 @@ pull_archive_api <- function(name, id, config, remote) {
...
+      if (!(name %in% valid$name)) {
+        stop(sprintf("No versions of report '%s' found at '%s'",
+                     name, remote$name),
+             call. = FALSE)
+      }
+    })
+
   if (id == "latest") {
     ## Resolve id
     v <- montagu::montagu_reports_report_versions(name, remote)
```
JamesThompson1729: Shouldn't this be the same as `v` on line 10? Is there something subtle going on with `withCallingHandlers`?

richfitz: you're completely right - that was redundant

PR #29, opened 2018-06-18
I1871

JamesThompson1729:
I'm not sure if we should be logging missing packages (lines 512-513), or just immediately stop with an error meesage

### DIFF
--- a/R/recipe_run.R
+++ b/R/recipe_run.R
@@ -507,6 +507,14 @@ orderly_prepare_data <- function(config, info, parameters, envir) {
   } else {
     hash_resources <- NULL
   }
+  missing_packages <- setdiff(info$packages, .packages(TRUE))
+  
+  if (length(missing_packages) > 0) {
+    orderly_log("packages",
+                sprintf("%s", missing_packages)) 
+    stop(paste("Missing packages:", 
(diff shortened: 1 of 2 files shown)


R/recipe_run.R
```diff
@@ -507,6 +507,14 @@ orderly_prepare_data <- function(config, info, parameters, envir) {
   } else {
     hash_resources <- NULL
   }
+  missing_packages <- setdiff(info$packages, .packages(TRUE))
+  
+  if (length(missing_packages) > 0) {
+    orderly_log("packages",
```
richfitz: I don't think we should log here, just error

R/recipe_run.R
```diff
@@ -507,6 +507,14 @@ orderly_prepare_data <- function(config, info, parameters, envir) {
...
+                sprintf("%s", missing_packages)) 
+    stop(paste("Missing packages:", 
+               paste(missing_packages, collapse = ", ")))
```
richfitz: I think that most errors `squote` the arguments (so here this would be `paste(squote(missing_packages, collapse = ", "))` for slightly nicer printing (this will affect the strings in your tests)

PR #37, opened 2018-07-19
i1930: new orderly schema

richfitz:
Getting this one in before it gets any bigger!

The actual schema can be seen here: https://vimc.github.io/orderly/schema/relationships.html - see also `inst/database/schema.yml`

The approach I have taken is to keep all the old db code in here so that we can merge this without having to get everything through the api all done in one go.  So there will be a second change that removes the old version later.

I've wired up "version 2" code to be called from the "version 1"  functions (init, rebuild and commit).

There's quite a bit going on in here (sorry) and I've resisted the urge to do much further refactoring.


R/db.R
```diff
@@ -128,40 +132,17 @@ report_db_cols <- function() {
...
 ##' @export
-orderly_rebuild <- function(config = NULL, locate = TRUE) {
+orderly_rebuild <- function(config = NULL, locate = TRUE, verbose = TRUE) {
   config <- orderly_config_get(config, locate)
-  report_db_rebuild(config)
+  report_db_rebuild(config, verbose)
   invisible(NULL)
 }
 
-## Need to merge this in with the above bits; need to merge the custom
```
richfitz: this was just dead code

inst/database/schema.yml
```diff
@@ -0,0 +1,148 @@
...
+# We could go more detailed here and get the report name and id but
+# that would duplicate information so I'm not doing that here.  (The
+# information redundant in the yml/rds but that's ok I feel).
+#
+# This is a bit of a risk because it is going to assume that the
+# artefact really was included in the upstream report.  But we can
+# depend on _any_ file not just dependencies, so this might fail for
+# montagu-reports which would make this inviable.  That should be
+# tightened up a lot in the recipe_read!
+depends:
```
JamesThompson1729: We probably need a `pinned` vs `use latest` flag here to indicate that the report really, truly did want to use this version of this report and not just that this was the latest version at the time

inst/database/schema.yml
```diff
@@ -0,0 +1,148 @@
...
+# Enum table for types of file use
+file_use:
+  columns:
+    - name: {type: TEXT}
+  values:
+    - {name: source}
+    - {name: script}
+    - {name: resource}
+    - {name: orderly_yml}
+    - {name: dependency}
```
MartinEden: Not using this any more?

inst/database/schema.yml
```diff
@@ -0,0 +1,148 @@
...
+# useful starting place for deduplicating the orderly archive in a
+# fairly safe way.  I don't think that we want to (necessarily) try
+# and index all files but certainly key inputs and outputs.
+file:
+  columns:
+    - hash: {type: TEXT}
+    - size: {type: BIGINT}
+
+# Enum table for types of file use
+file_use:
```
MartinEden: `file_purpose`

PR #39, opened 2018-07-27
i2008

richfitz:
Orderly migration

This doesn't include _all_ of the migrations required with the changes on i1930 (#37); we'll do them there


R/testing.R
```diff
@@ -119,6 +119,13 @@ demo_change_time <- function(id, time, path) {
   dat$id <- id_new
   writeLines(yaml::as.yaml(dat), yml)
 
+  rds <- path_orderly_run_rds(p)
```
richfitz: Most of this code is in 154e9da which can also be found on the i1930/#37 branch

PR #47, opened 2018-09-18
I1756

JamesThompson1729:
Functionality for a global resource directory.
* Create a directory `<DIR_NAME>` in the orderly root
* Add a `global_resources` parameter to with value `<DIR_NAME>` to `orderly_config.yml`


R/recipe_run.R
```diff
@@ -381,6 +381,25 @@ recipe_prepare_workdir <- function(info) {
...
+    root_path <- normalizePath(config$path, mustWork = TRUE)
+    global_resource_dir <- file.path(root_path, config$global_resources)
+
+    if (!dir.exists(global_resource_dir))
+      stop(sprintf("Global resource folder does not exist (expected: %s)",
+                   config$global_resources))
+
+    path_global_recs <- file.path(global_resource_dir, info$global_resources)
+    dest_resources_src <- file.path(info$workdir, info$global_resources)
+    i <- is_directory(path_global_recs)
```
richfitz: I _think_ this whole section can be simplified as

```r
file_copy(path_global_recs, info$workdir, recursive = TRUE)
```

R/recipe_run.R
```diff
@@ -381,6 +381,25 @@ recipe_prepare_workdir <- function(info) {
     orderly_log("depends", str)
     file_copy(src, dst)
   }
+  
+  if (!is.null(info$global_resources)) {
```
richfitz: This needs moving about:

* the check that the `global_resources` entry exists and is a directory belongs in the config (see `orderly_config_read_yaml`)
* the there needs to be a check here that all the global resources in _this_ orderly report actually exist (that check is missing at the moment I believe)

tests/testthat/test-global.R
```diff
@@ -0,0 +1,46 @@
...
+    orderly_run("example", config = path, id_file = tmp, echo = FALSE),
+    NA # expect no errors
+  )
+})
+
+test_that("no global folder", {
+  path <- prepare_orderly_example("global")
+  # now we break the orderly_config.yml
+  path_config <- file.path(path, "orderly_config.yml")
+  config_lines = readLines(path_config, -1)
```
richfitz: arrow, not `=`, and no need for `-1`

tests/testthat/test-global.R
```diff
@@ -0,0 +1,46 @@
...
+  config_lines[6] <- "  invalid_path"
+  writeLines(config_lines, path_config)
+  
+  tmp <- tempfile()
+  expect_error(
+     orderly_run("example", config = path, id_file = tmp, echo = FALSE),
+     paste("Global resource folder does not exist", 
+           "(expected: invalid_path)"),
+     fixed = TRUE
+     )
```
richfitz: paren belongs on previous line

tests/testthat/test-global.R
```diff
@@ -0,0 +1,46 @@
...
+})
+
+test_that("missing global file", {
+  path <- prepare_orderly_example("global")
+  # now we break the report yaml
+  path_example <- file.path(path, "src", "example")
+  path_yaml <- file.path(path_example, "orderly.yml")
+  # the final line of the yaml is globa file, so we change that
+  config_lines = readLines(path_yaml, -1)
```
richfitz: change as above

R/recipe_run.R
```diff
@@ -381,6 +381,25 @@ recipe_prepare_workdir <- function(info) {
...
+    # check the global resource folder exist
+    root_path <- normalizePath(config$path, mustWork = TRUE)
```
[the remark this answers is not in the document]
JamesThompson1729: We have to duplicate line 348 since the value of `src` can be overwritten on line 370.
There might be a nicer way to do this.

R/recipe_run.R
```diff
@@ -381,6 +381,25 @@ recipe_prepare_workdir <- function(info) {
...
+    global_resource_dir <- file.path(root_path, config$global_resources)
+
+    if (!dir.exists(global_resource_dir))
```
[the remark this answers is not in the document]
JamesThompson1729: I'm not sure the behaviour is identical, I will investigate.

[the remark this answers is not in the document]
JamesThompson1729: is_directory doesn't check for existence and returns NA if the file/dir doesn't exist.
dir.exists returns false is the file/dir doesn't exist. 

I don't know if we're relying on is_directory returning NA anywhere.
(I'm not sure about dir.exists behaviour on non-UNIX OSs)

R/config.R
```diff
@@ -61,6 +61,13 @@ orderly_config_read_yaml <- function(filename, path) {
       match_value(api_server_identity, names(info$api_server))
   }
 
+  if (!is.null(info$global_resources)) {
+    global_dir <- file.path(path, info$global_resources)
+    if(!dir.exists(global_dir)) {
```
richfitz: `orderly:::assert_is_directory`

PR #53, opened 2018-11-06
i1503: orderly changelog

richfitz:
The beginnings of orderly changelog support!

The workflow and some rationale is explained in `vignettes/changelog.Rmd` - please suggest additions if there are extra bits to add.

The good news is that the changes are somewhat isolated:

* inst/database/schema.yml: one new table
* R/db2.R: read in an already processed changelog and add it to the table
* R/paths.R: two new paths
* R/recipe_run.R: various tweaks to the logic around `recipe_prepare` / `recipe_run` mostly as the message processing moves to become part of the changelog

The code in R/changelog.R is all brand new so any changes to logic can be done, really

Possible things to tighten up

1. we never check the changelog.json against the provided schema (I've done this manually though). Doing so would add a `Suggests:` dependency on `jsonvalidate`, but that's not a huge deal
2. changelog.json should be forbidden within `src`, but this applies more generally and could warrant another issue


richfitz:
Hmm, perhaps valid labels should go into the `orderly_config.yml`?

richfitz:
Closed in favour of #54 in order to preserve branch names

PR #59, opened 2018-11-12
i2374: Don't redocument on build

richfitz:
This somehow has worked:
http://teamcity.montagu.dide.ic.ac.uk:8111/viewLog.html?buildId=26494&buildTypeId=montagu_Orderly_BuildContainer&tab=buildLog&_focus=502#_state=502

but I don't actually see how. Because we run the container with
user permissions but copy the sources in with root, the document
step that pkgdown runs can't possibly succeed.

I am also opportunistically clearing out the docs directory before
build, which will mean that deleted files are not included

### DIFF
--- a/scripts/build_website
+++ b/scripts/build_website
@@ -61,6 +61,7 @@ docker run --rm \
        $ORDERLY_DEV \
        ./create_schema
 
+rm -rf $DOCS_DIR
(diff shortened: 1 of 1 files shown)


richfitz:
http://teamcity.montagu.dide.ic.ac.uk:8111/viewLog.html?buildId=26585&

PR #70, opened 2019-01-15
i2453: separate out montagu from orderly

richfitz:
This requires some thoughts.  I'm fairly certain that I have broken one or two deploy related things, so we'll need to be careful rolling this out (in particular - does the configuration updates in https://github.com/vimc/montagu-reports PR flow through so that the appropriate server options will make it through ok?)


R/remote.R
```diff
@@ -242,83 +201,74 @@ get_default_remote <- function(config = NULL, locate = TRUE) {
...
 
-remote_name <- function(remote) {
-  check_remote_type(remote)
-  if (inherits(remote, "montagu_server")) {
-    remote$name
-  } else if (inherits(remote, "orderly_remote_path")) {
-    as.character(remote)
-  }
+## Test to see if something (might) implement our remote interface
+implements_remote <- function(x) {
```
JamesThompson1729: When are we expecting to have to deal with objects that aren't the defined R6 classes?
Isn't `x$class == orderly_remote_path` safer?

R/remote_path.R
```diff
@@ -1,36 +1,57 @@
...
+
+    initialize = function(path, name) {
+      assert_file_exists(path)
+      path <- normalizePath(path, "/", mustWork = TRUE)
+      if (!file.exists(path_orderly_config_yml(path))) {
+        stop("Does not look like an orderly repository: ", squote(path))
+      }
+      self$config <- orderly_config(path)
+      self$name <- name %||% self$config$path
+      lockBinding(quote(config), self)
```
JamesThompson1729: I think lockBinding is (more or less) equivalent to making a member variable `const`.
I don't understand what `quote` is doing here though.

R/remote_path.R
```diff
@@ -1,36 +1,57 @@
...
+      lockBinding(quote(name), self)
+    },
+
+    list = function() {
```
JamesThompson1729: Is `list_reports` clearer? Having functions with the same name as built in primitives makes me anxious.

R/remote_path.R
```diff
@@ -1,36 +1,57 @@
...
+      orderly_list(self$config)
+    },
+
+    list_versions = function(name) {
+      d <- orderly_list_archive(self$config)
+      d$id[d$name == name]
+    },
+
+    pull = function(name, id, root) {
```
JamesThompson1729: Can pull and push ever fail here? Do produce messages or errors somewhere else?

R/remote.R
richfitz: See the object created in https://github.com/vimc/montagu-r/pull/13 - that will confirm to the specification here but not have that class name.

In an ideal world R6 would have the concept of an interface (see, e.g., https://github.com/r-lib/R6/issues/163, https://github.com/r-lib/R6/issues/82). This function tries to bodge around that to see if we have all the methods that are expected. Ideally we'd check more about the function signature too but it's a bit dull to do.

But the core idea is that `orderly` defines what is needed to be supplied and the other packages can implement things that `orderly` does not know about without having to modify `orderly` itself - I think in OOP circles this is the [Open-Closed Principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle)

R/remote_path.R
richfitz: you are right about `lockBinding`. Quote just generates a *symbol* which is what `lockBinding` used to require (now it seems that a string is ok and it will use `as.name` internally to create a symbol)

R/remote_path.R
richfitz: that would be less ambiguous with `list_versions`.

There's no danger of colliding with internal functions though (they're not keywords in R) because this is always called `thing$list()` which is equivalent to `thing[["list"]]()`

R/remote_path.R
richfitz: these can absolutely fail (in fact `push` always fails for the api based remote driver) but that will always propagate out ok

PR #71, opened 2019-01-17
I2384

JamesThompson1729:
Asks a user if they want to install missing packages.
Prints the message
```
Should I try to install missing packages by running:

    install.packages(c('foo','bar'')) 
```

Stuff I'm not sure it right
* This is disabled when logging is turn off, in this case it automatically halts with an error.
* `orderly_log` is used with `INSTALL` topic for the installation of each package (this is probably overkill).
* There doesn't seem a nice way to test this with `test_that`, so I've disabled it and just checked the error


R/util.R
```diff
@@ -610,3 +610,46 @@ abbreviate <- function(x, len = round(getOption("width", 80) * 0.8)) {
...
+handle_missing_packages <- function(missing_packages) {
+  ## collapse vector to packages to string "c('pckg_1','pckg_2')"
+  vector_packages <- sprintf("install.packages(c(%s))",
+                             paste(sprintf("'%s'", missing_packages),
+                                   collapse=","))
+
+  ## check if we are interactive and logging is active...
+  logging <- !isTRUE(getOption("orderly.nolog"))
+  if (interactive() && logging) {
+    ## ...if so ask if Orderly should try to install the pacakges
```
richfitz: https://codecov.io/gh/vimc/orderly/src/cda69847a1a5ae21365d5b7761adc49ae3900712/R/util.R

Not sure why codecov's bot has not commented on the PR but this section here is untested.  This is the sort of thing that is a nightmare to test though...

R/util.R
```diff
@@ -610,3 +610,46 @@ abbreviate <- function(x, len = round(getOption("width", 80) * 0.8)) {
   }
   x
 }
+
+handle_missing_packages <- function(missing_packages) {
+  ## collapse vector to packages to string "c('pckg_1','pckg_2')"
+  vector_packages <- sprintf("install.packages(c(%s))",
```
richfitz: we have an `squote` function for this job already - please use:

```
sprintf("install.packages(c(%s))", paste(squote(missing_packages), collapse = ", "))
``` 

(note space after comma in the `collapse`)

R/util.R
richfitz: please pull all the TRUE branch of this clause into its own function (`install_missing_packages` perhaps). Then we can use the mocking package `mockery` to start testing branches in it more easily

R/util.R
```diff
@@ -610,3 +610,46 @@ abbreviate <- function(x, len = round(getOption("width", 80) * 0.8)) {
...
+                             paste(sprintf("'%s'", missing_packages),
+                                   collapse=","))
+
+  ## check if we are interactive and logging is active...
+  logging <- !isTRUE(getOption("orderly.nolog"))
+  if (interactive() && logging) {
+    ## ...if so ask if Orderly should try to install the pacakges
+    question <- "Should I try to install missing packages by running:"
+    install_command <- paste0("\n", question, "\n\n",
```
richfitz: please rewrite using `sprintf`

R/util.R
```diff
@@ -610,3 +610,46 @@ abbreviate <- function(x, len = round(getOption("width", 80) * 0.8)) {
...
+                              "    ", vector_packages)
+
+    try_install <- (utils::menu(c("no", "yes"), FALSE,
+                               title = install_command) == 2)
+
+    if (try_install) {
+      for (pckg in missing_packages) {
```
richfitz: packages should be installed as a vector - this is different to what the function advertises that it will do

R/util.R
```diff
@@ -610,3 +610,46 @@ abbreviate <- function(x, len = round(getOption("width", 80) * 0.8)) {
...
+        orderly_log("INSTALL",
+                    sprintf("Attempting to install %s", pckg))
+        withCallingHandlers(install.packages(pckg, quiet = TRUE),
```
richfitz: less errorprone (because `install.packages` is almost impossible to work around) given we know that these are not currently installed, is to just check that after running the install they are found.

R/util.R
```diff
@@ -610,3 +610,46 @@ abbreviate <- function(x, len = round(getOption("width", 80) * 0.8)) {
...
+    } else {
+      stop(paste("Missing packages:",
+                  paste(squote(missing_packages), collapse = ", ")))
+    }
+  } else {
+    ## ...we're not in interactive environment so just print out the command
+    question <- "To install the missing packages run:"
+    install_command <- paste0("\n", question, "\n\n",
+                              "    ", vector_packages)
+    stop(paste("Missing packages:",
```
richfitz: please use `sprintf` in the statements above and below to construct strings rather than `paste`

tests/testthat/test-orderly.R
```diff
@@ -112,6 +112,8 @@ test_that("orderly_run_info: is_latest detects latest version", {
 
 
 test_that("orderly_test_start failure resets working directory", {
+  # if logging is on this test will wait for user input, so turn off
+  orderly_log_off()
```
richfitz: when changing global state in tests, use `on.exit` to restore

It might be better though to have a further option to disable package installation too (`orderly.no_install_missing`)

R/util.R
```diff
@@ -610,3 +610,49 @@ abbreviate <- function(x, len = round(getOption("width", 80) * 0.8)) {
   }
   x
 }
+
+handle_missing_packages <- function(missing_packages) {
+  ## check if we are interactive and logging is active...
+  logging <- !isTRUE(getOption("orderly.nolog"))
+  if (interactive() && logging) {
+    install_missing_packages(missing_packages)
```
richfitz: in order to make this testable, can you try adding an argument `force = FALSE` to `handle_missing_packages` then we can call `handle_missing_packages(packages, TRUE)` and trigger this line which remains untouched in tests

R/util.R
```diff
@@ -611,3 +611,59 @@ abbreviate <- function(x, len = round(getOption("width", 80) * 0.8)) {
...
+  ## collapse vector to packages to string "c('pckg_1','pckg_2')"
+  vector_packages <- sprintf("install.packages(c(%s))",
+                             paste(squote(missing_packages), collapse = ", "))
+
+  ## ...if so ask if Orderly should try to install the pacakges
+  question <- "Should I try to install missing packages by running:"
+  install_command <- sprintf("\n%s\n\n    %s", question, vector_packages)
+
+  if (prompt_ask_yes_no(install_command)) {
+      install_packages(missing_packages)
```
richfitz: indenting is off by two

R/util.R
```diff
@@ -611,3 +611,59 @@ abbreviate <- function(x, len = round(getOption("width", 80) * 0.8)) {
...
+  }
+}
+
+install_missing_packages <- function(missing_packages) {
+  ## collapse vector to packages to string "c('pckg_1','pckg_2')"
+  vector_packages <- sprintf("install.packages(c(%s))",
+                             paste(squote(missing_packages), collapse = ", "))
+
+  ## ...if so ask if Orderly should try to install the pacakges
+  question <- "Should I try to install missing packages by running:"
```
richfitz: This looks like it duplicates the code on line 623-625

R/util.R
```diff
@@ -611,3 +611,59 @@ abbreviate <- function(x, len = round(getOption("width", 80) * 0.8)) {
...
+                 paste(squote(missing_packages), collapse = ", ")))
+  }
+}
+
+prompt_ask_yes_no <- function(prompt) {
+  return(utils::menu(c("no", "yes"), FALSE, title = prompt) == 2)
+}
+
+show_question <- function() {
+  return(interactive() && !isTRUE(getOption("orderly.nolog")))
```
richfitz: no explicit `return` statement here or in `prompt_ask_yes_no` please

R/util.R
```diff
@@ -611,3 +611,59 @@ abbreviate <- function(x, len = round(getOption("width", 80) * 0.8)) {
...
+}
+
+install_packages <- function(missing_packages) {
+  ## try to install missing packages...
+  ## warnings are down graded to messages since any problem should be
+  ## immediately caught on the next line
+  install.packages(missing_packages, quiet = TRUE)
+  ## ...then check that they have been sucessful
+  found_packages <- missing_packages %in% rownames(installed.packages())
```
richfitz: this will be easier with `setdiff`:

```
msg <- setdiff(missing_packages, rownames(installed.packages())
if (length(msg) > 0) {
  stop(sprintf("Could not install these packages: %s",
               paste(squote(msg, collapse = ", ")))
}
```

PR #72, opened 2019-01-18
I2137

JamesThompson1729:
This PR does two things:

* Moves the resource hashing to where the resources are copied and then checks the resources haven't changed after the report is run
* Adds resource size to the information stored in the orderly field in the database.

This will almost certainly fail some appveyor tests since:

* The shiny reports quietly define a directory as a resource and then add file to that directory
* The new column in the database does not exist in the existing tests.

I am creating the PR now as I am not sure what is the correct way to deal with these issues.


R/recipe_run.R
```diff
@@ -237,6 +238,17 @@ recipe_run <- function(info, parameters, envir, config, echo = TRUE) {
   hash_data_rds <- con_rds$mset(prep$data)
   stopifnot(identical(hash_data_csv, hash_data_rds))
 
+  ## make sure the resources have the same hashes (and size?)
+  post_run_resources <- get_resource_info(info)
+  stopifnot(identical(resource_info$hash_resources,
```
richfitz: These are not going to be clear enough - `stopifnot` should only be used for internal assertions and not for user facing errors.  Replace with an if block here, and explicitly test it (see test file for what needs doing there).  We'll back out the size changes for now, but in general if the size has changed the hash will change so that is sufficient.

R/db.R
```diff
@@ -131,6 +131,7 @@ report_db_cols <- function() {
     hash_orderly = "TEXT",
     hash_input = "TEXT",
     hash_resources = "TEXT", # should be json (dict)
+    size_resources = "TEXT", # should be json (dict)
```
richfitz: The db.R file is legacy and all changes will need to go in db2.R and into inst/database/schema.yml.  Can you back these changes out for now please and we'll do that on a separate ticket

R/recipe_run.R
```diff
@@ -514,30 +538,45 @@ orderly_prepare_data <- function(config, info, parameters, envir) {
...
   n_dev <- length(grDevices::dev.list())
 
-  hash_resources <- hash_files(expand_directory_list(info$resources))
-  if (length(hash_resources) > 0L) {
-    orderly_log("resources",
-                sprintf("%s: %s", names(hash_resources), hash_resources))
-  } else {
-    hash_resources <- NULL
-  }
+  # hash_resources <- hash_files(expand_directory_list(info$resources))
```
richfitz: please remove the commented out sections. Your function is a nicer way of doing thi

R/recipe_run.R
```diff
@@ -553,6 +592,43 @@ orderly_prepare_data <- function(config, info, parameters, envir) {
...
+    size_global <- file_size(expand_directory_list(info$global_resources))
+  } else {
+    size_global <- NULL
+  }
+
+  ret <- list(hash_resources = hash_resources,
+              hash_global = hash_global,
+              size_resources = size_resources, size_global = size_global)
+  
+#  print(ret)
```
richfitz: remove this commented section please

R/recipe_run.R
```diff
@@ -514,30 +538,45 @@ orderly_prepare_data <- function(config, info, parameters, envir) {
...
                paste(squote(missing_packages), collapse = ", ")))
   }
 
-  ret <- list(data = ldata, hash_resources = hash_resources,
-              hash_global = hash_global, n_dev = n_dev)
+  # ret <- list(data = ldata, hash_resources = hash_resources,
+  #             hash_global = hash_global, n_dev = n_dev,
+  #             size_resource = size_resource, size_global = size_global)
+  
+  ret <- list(data = ldata, n_dev = n_dev)
```
richfitz: we no longer return hash_resources and hash_global here? were they not being used?

R/util.R
```diff
@@ -159,6 +159,28 @@ hash_object <- function(object) {
   digest::digest(object)
 }
 
+file_size <- function(filenames, named = TRUE) {
```
richfitz: This can come out, but `file.size` will do this.  There's no point adding the error condition here though as this should be tested elsewhere (and it is I think) - the message as it stands would tie a generally named function to a specific task.  If a region of code is untestable it should not be here

tests/testthat/test-run.R
```diff
@@ -63,6 +63,7 @@ test_that("run", {
   expect_identical(unname(unlist(run$hash_artefacts, use.names = FALSE)),
                    hash_files(file.path(p, "mygraph.png"), FALSE))
   expect_identical(run$hash_resources, list())
+  expect_identical(run$size_resources, list())
```
richfitz: these will come out.

What needs to come in is a test where you create a script file that when run does modify the inputs (one test modifying a resource, then in another test modify a global resource) and check that your condition does trigger as expected

R/recipe_run.R
```diff
@@ -530,30 +568,43 @@ orderly_prepare_data <- function(config, info, parameters, envir) {
...
+  ## compare resource hashes here
+  found <- names(post_run_hashes)
+  expected <- names(pre_run_hashes)
 
-  if (!is.null(info$connection)) {
-    ret$con <- data[[info$connection]]
+  ## we expected a resource but can't find it.
+  not_found <- which(!(expected %in% found))
+  if (any(not_found)) {
+    stop("Script either did not copy or deleted resources: ",
```
richfitz: why "did not copy"? The script should not copy a resource

R/recipe_run.R
```diff
@@ -530,30 +568,43 @@ orderly_prepare_data <- function(config, info, parameters, envir) {
...
 
-  for (p in info$packages) {
-    library(p, character.only = TRUE)
-  }
-  for (s in info$sources) {
-    source(s, envir)
+  ## I'm not sure this can ever happen
+  ## Orderly has found a copied resource that was not expected...
+  not_expected <- which(!(found %in% expected))
+  if (any(not_expected)) {
```
richfitz: I don't understand this - would this not throw on any working files that are created - which we already have a system in place for checking for?  (See `unexpected` within this file)

richfitz: codecov says you never managed to trigger this either - I'm pretty sure we can remove this

R/recipe_run.R
JamesThompson1729: That should probably read Orderly has failed to copy across a resource OR the script has deleted a resources file.
The failure to copy is caught elsewhere so I'll change this to `Script deleted resources`

R/recipe_run.R
JamesThompson1729: I was working through the all the possible permutations of finding / not finding resources. This is  caught by unexpected.

PR #75, opened 2019-01-31
I2230

JamesThompson1729:
Logic on auto detecting README.md
* 1 Is `README.md` explicitly included as a resource in the report yml? If so copy it as a resource.
_if not_
* 2. Is there a `has_readme` boolean value in the report yml? If so try to copy `src/README.md`.
_If not_
* 3. Is there a file named `README.md` in `/src/`? If so copy `src/README.md`.


R/db2.R
```diff
@@ -322,23 +322,37 @@ report_data_import <- function(con, workdir, config) {
   ## TODO: patch this back in for the saved rds I think
   hash_orderly_yml <- hash_files(file.path(workdir, "orderly.yml"), FALSE)
 
+  ## there might be a better way to this.
+  ## At the moment we are repeatedly checking if a readme has been copied
+  ## maybe add a field to the run info saying that we've copied a README?
+  if (file_exists(file.path(workdir, "README.md"), check_case = FALSE)) {
+    readme_name <- "README.md"
```
richfitz: given the case-insensitive change above, should this not find the canonical case? I think you can do this with

```r
readme_name <- file_canonical_case("README.md")
```

but you should pair that with a test that includes something like `readme.md`

inst/database/schema.yml
```diff
@@ -71,6 +71,7 @@ file_purpose:
     - {name: resource}
     - {name: orderly_yml}
     - {name: global}
+    - {name: readme}
```
richfitz: can you bump the `ORDERLY_SCHEMA_VERSION` in `db2.R` please?

R/db2.R
JamesThompson1729: After a bit of investigation, `file_canonical_case` behaves oddly with test_that.
It looks like `.` is the test_that directory, so `file_canonical_case` will always return `NA`.

R/db2.R
```diff
@@ -322,23 +322,41 @@ report_data_import <- function(con, workdir, config) {
   ## TODO: patch this back in for the saved rds I think
   hash_orderly_yml <- hash_files(file.path(workdir, "orderly.yml"), FALSE)
 
+  ## there might be a better way to this.
+  ## At the moment we are repeatedly checking if a readme has been copied
+  ## maybe add a field to the run info saying that we've copied a README?
+  if (file_exists(file.path(workdir, "README.md"), check_case = FALSE)) {
+    if (file_has_canonical_case("README.md")) {
+      readme_name <- file_canonical_case("README.md")
```
richfitz: This line is never hit in the tests - can we simplify the logic here perhaps?

JamesThompson1729: What I've done here is wrong I think.
The canonical_case (via `file_split_base`) needs the full absolute path, so this will always fail to find a canonical form. If we pass in the full path e.g. `file_canonical_case(file.path(workdir, "README.md"))` then this breaks elsewhere since internally, Orderly uses relative paths.

I'll have a think about the best way forward.

R/db2.R
```diff
@@ -6,7 +6,7 @@
 ## namespace/module feature so that implementation details can be
 ## hidden away a bit further.
 
-ORDERLY_SCHEMA_VERSION <- "0.0.4"
+ORDERLY_SCHEMA_VERSION <- "0.0.5"
```
richfitz: Can you increase this to 0.0.6 and the package version to 0.5.17 to leap over the current versions? (don't forget to update the news version too)

PR #78, opened 2019-02-18
i2601: add public changelog version

richfitz:


R/db2.R
```diff
@@ -459,9 +470,46 @@ report_db2_destroy <- function(con) {
...
+  for (i in seq_len(nrow(dat))) {
+    j <- dat$report_version[[i]] <= published
+    if (!any(j)) {
+      break
+    }
+    p[[i]] <- published[[min(which(j))]]
+  }
+
+  prev <- dat$report_version_public
+  update <- is.na(p) != is.na(prev) | (!is.na(p) & !is.na(prev) & p != prev)
```
JamesThompson1729: I don't follow the the logic here. Unless R allowed to short circuit expressions like this, I think this should evaluate to `update <- is.na(p)`.

And I'm not 100% sure that `&` and `|` are higher precedence than `!=`.

There is probably a clearer way of expressing the logic in this line.

R/db2.R
```diff
@@ -459,9 +470,46 @@ report_db2_destroy <- function(con) {
...
+               "  FROM changelog",
+               "  JOIN report_version",
+               "    ON report_version.id = changelog.report_version",
+               "  JOIN changelog_label",
+               "    ON changelog_label.id = changelog.label",
+               " WHERE report_version.report = $1",
+               "   AND changelog_label.public",
+               " ORDER BY report_version")
+  dat <- DBI::dbGetQuery(con, sql, list(name))
+  dat$published <- dat$published == 1
```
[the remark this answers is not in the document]
richfitz: SQLite stores booleans as integers - it's very annoying. This line just sanitises the value so that we can treat it as the logical it really is

R/db2.R
richfitz: There is no short circuiting in here, but the logic does not expand as you suggest.  `!=` binds more tightly than `&`

The logic here is the following

|old|new|updated?|
|---|---|---|
|NA|NA|FALSE|
|x|x|FALSE|
|NA|x|TRUE|
|x|NA|TRUE|
|x|y|TRUE|

so we might be able to do it with:

```r
!(is.na(p) & is.na(prev) | (!is.na(p) & !is.na(prev) & p == prev))
```

but that's about as much logic.  What do you think?

PR #81, opened 2019-02-19
i2642: Prevent git reference switching on production

richfitz:
This honours a configuration already present in montagu-reports, but currently ignored by orderly

### DIFF
--- a/R/runner.R
+++ b/R/runner.R
@@ -4,9 +4,15 @@
 ##'
 ##' @title Orderly runner
 ##' @param path Path to use
-##' @param allow_ref Allow git to change branches/ref for run
+##'
+##' @param allow_ref Allow git to change branches/ref for run.  If not
+##'   given, then we will look to see if the orderly configuration
+##'   disallows branch changes (based on the
+##'   \code{ORDERLY_API_SERVER_IDENTITY} environment variable and the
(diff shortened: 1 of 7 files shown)


R/runner.R
```diff
@@ -4,9 +4,15 @@
...
+##'
+##' @param allow_ref Allow git to change branches/ref for run.  If not
+##'   given, then we will look to see if the orderly configuration
+##'   disallows branch changes (based on the
+##'   \code{ORDERLY_API_SERVER_IDENTITY} environment variable and the
+##'   \code{master_only} setting of the relevant server block.
+##'
 ##' @export
-orderly_runner <- function(path, allow_ref = TRUE) {
+orderly_runner <- function(path, allow_ref = NULL) {
```
JamesThompson1729: if a user creates an `orderly_runner` with `allow_ref = TRUE`, we ignore the `master_only` in the `orderly_config`.

So a user could still run a report from a non-master branch on a remote wirh `master_only`, (although it would be much more difficult to do accidentally).

is this behaviour we want or should the `master_only` option override the `allow_ref` argument?

richfitz: You are correct. I'm not sure what the Right Thing to do here is but we will deprecate the argument soon I think

PR #82, opened 2019-02-20
i2517: Better behaviour with report timeout

richfitz:
This PR does very little towards the ticket, just updating a few arguments as they're passed through to the runner.

We have a new `wait` parameter: that's the _client-side_ timeout as opposed to `timeout` which is for the server.  The PR in `montagu-r` will forward that appropriately

See https://github.com/vimc/montagu-r/pull/20 and https://github.com/vimc/orderly.server/pull/10

### DIFF
--- a/R/remote.R
+++ b/R/remote.R
@@ -114,7 +114,13 @@ push_archive <- function(name, id = "latest", config = NULL, locate = TRUE,
 ##'
 ##' @param parameters Parameters for the reprt
 ##'
-##' @param timeout Time to wait for the report to be returned (in seconds)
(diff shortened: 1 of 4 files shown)


R/remote.R
```diff
@@ -136,14 +142,14 @@ push_archive <- function(name, id = "latest", config = NULL, locate = TRUE,
 ##'
 ##' @export
 orderly_run_remote <- function(name, parameters = NULL, ref = NULL,
```
JamesThompson1729: We don't pass `wait` to `remote$run`

PR #87, opened 2019-04-29
VIMC-2795: Support multiple databases

richfitz:
This PR will change orderly's handling of databases to allow more than one (and also zero) databases to be supported in configuration and used in reports.

This _looks_ like a big PR but it's not that bad:

```
$ git diff  --stat master.. -- R
 R/config.R        |  36 ++++++++++++-----
 R/db.R            |  36 +++++++++++------
 R/db2.R           |  17 +++-----
 R/logging.R       |   9 +++++
 R/migrate.R       |   5 +++
 R/recipe_commit.R |   7 +++-
 R/recipe_read.R   | 116 ++++++++++++++++++++++++++++++++++++++++++------------
 R/recipe_run.R    |  53 +++++++++++++++++++------
 R/testing.R       |  10 +++--
 R/util.R          |  14 +++++--
 10 files changed, 225 insertions(+), 78 deletions(-)
```

and

```
$ git diff  --stat master.. -- inst
...
 64 files changed, 246 insertions(+), 78 deletions(-)
```

and

```
$ git diff  --stat master.. -- tests/
 tests/testthat/create-reference.R         |   8 +++++
 tests/testthat/example/orderly_config.yml |  15 +++++-----
 tests/testthat/example_config.yml         |   7 +++--
 tests/testthat/example_report.yml         |  24 ++++++++-------
 tests/testthat/helper-orderly.R           |  10 +++++++
 tests/testthat/reference/0.5.18.zip       | Bin 0 -> 975960 bytes
 tests/testthat/test-config.R              |  78 +++++++++++++++++++++++++++++++++++-------------
 tests/testthat/test-db.R                  |   2 +-
 tests/testthat/test-envir.R               |  15 +++++-----
 tests/testthat/test-main.R                |   6 ++++
 tests/testthat/test-migrate.R             |  36 ++++++++++++++++++++++
 tests/testthat/test-readme.R              |   3 +-
 tests/testthat/test-recipe-read.R         | 136 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++-----
 tests/testthat/test-run.R                 |  62 +++++++++++++++++++++++++++++++++++---
 tests/testthat/test-z-runner.R            |   7 +++--
 15 files changed, 344 insertions(+), 65 deletions(-)
```

i.e., there is relatively little code change  but these have a fairly fundamental efffect on things like the tests and examples and that blows out the number of files changed and the number of changes that are not that interesting.

---

This is a fairly major change to the configuration and we should probably have thought about this more carefully at the outset.  Up to this point we have assumed that there is a single database that *must* be configured.  After this change there might be zero or more databases.  The query interface will be much more general and this should leave us in a more flexible position to do things like set options around queries or even support non-SQL databases.

There are a number of places where it is implicit that we have one and only one database connection and these all need finessing to deal with multiple databases.  As much as possible I want to support backward compatibility with previous uses, and as such we'll leave most of the configuration in examples untouched I hope.

In the `orderly.yml` for each report:

`data`: - up to now we use

```
data:
  r_object_name: <sql query>
```

with multiple queries being allowed and mapped to different R objects.  We'll keep support for this format, but prefer (and document)

```
data:
  r_object_name:
    query: <sql query>
    database: <dbname>
```

with the dbname being one of the names in the database configuration

`connection`: - up to now we use 

```yaml
connection: name_of_connection
```

We'll support that with an error if there is more than one db.  If there is more than one database then we'll use

```yaml
connection:
  db1: name_of_connection1
  db2: name_of_connection2
```

and of course this will work for the single db case.  It's not clear if this *absolutely* requires db access.  At the moment we store a boolean `connection` against the report version - we could interpret this as being any connection was used with complete compatibility with previous versions and migrate that later if we want more details.

Similar support was required for views but these turned out to be poorly tested, and I have expanded that testing somewhat.  These are now also included in the demo.

The documentation and report templates have been adjusted.

Lots of changes to tests.

Old behaviour is supported by a series of warnings, which are also tested

A migration updates old versions that existed before this concept to use the database name `source`.


R/recipe_commit.R
```diff
@@ -51,9 +51,14 @@ recipe_commit <- function(workdir, config) {
   ## Copy the _files_ over, but we'll roll this back if anything fails
   dest <- copy_report(workdir, dat$name, config)
 
+  DBI::dbBegin(con)
```
richfitz: These changes are unrelated to the PR but I ran into a point where this should have been done - only commit the updates to the DB if all updates work

richfitz:
> I intuitively expected to be able to type production instead of (or as an alternative to) source in the various database - but the error message makes it immediately clear what to do. (I guess if we wanted to allow production to be named, that would be in the orderly_config.yml on montagu-reports anyway.

This is a good point, somewhat addressed here:

https://github.com/vimc/montagu-reports/pull/310/files

The issue is that we would have to write a one-off migration to go through all the reports and patch `source` to be `production` after applying the main migration in this PR

PR #97, opened 2019-05-21
VIMC-2882: back up database periodically

richfitz:
This PR will add a periodic backup to be run from the runner.

The use case/issue is that if we are backing up the system with rsync we have the potential to back up an inconsistent db, which is a problem with data now being stored on the OrderlyWeb side as well as the rebuildable orderly db.

With the changes here the db is backed up every 10 minutes by default.  To make the backup process safe to inconsistencies, we first backup the backup by moving it to `<name>.prev` (which is atomic) then doing the new backup

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: orderly
 Title: Lightweight Reproducible Reporting
-Version: 0.6.4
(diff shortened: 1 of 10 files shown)


richfitz:
(rebased to incorporate #96)

PR #99, opened 2019-06-03
VIMC-2696: Global PK for changelog entries

richfitz:
This is needed for orderly web because we need something that will persist across DB rebuilds.

Before merging we should check to see the impact on the reporting portal and orderly web

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: orderly
 Title: Lightweight Reproducible Reporting
-Version: 0.6.6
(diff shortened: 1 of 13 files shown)


richfitz:
As discussed in the office today, we now include a sortable integer field in the changelog table (`ordering`) that can be used to get the insertion order.  So if we replace previous uses of the `id` column with `ordering` this should provide a drop-in replacement (can we look at what is needed further on Tuesday @EmmaLRussell?)

PR #105, opened 2019-06-17
VIMC-2931: add examples to exported orderly functions

richfitz:
This PR adds fairly basic examples to orderly functions.  We could probably do with more in places, but it provides a starting point.

Unwritten CRAN rules require examples for all exported functions, so this is a prereq for CRAN release (VIMC-2421)

(There is very little code change in this PR, despite the apparent size)


R/logging.R
```diff
@@ -17,21 +25,60 @@
 ##' }
```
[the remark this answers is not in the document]
richfitz: this is useful, thanks - looks like that has never made sense (and therefore by extension nobody has ever used that bit of the docs!)

R/remote.R
```diff
@@ -164,20 +179,53 @@ orderly_run_remote <- function(name, parameters = NULL, ref = NULL,
...
-##' @param value A string describing a remote, or \code{NULL} to clear
+##'
+##' @param value A string describing a remote, a remote object, or
+##'   \code{NULL} to clear
+##'
 ##' @inheritParams orderly_list
 ##' @export
 ##' @rdname orderly_default_remote
+##' @examples
+##' # Same setup as in ?orderly_remote_path, with a remote orderly:
```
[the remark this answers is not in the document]
richfitz: If you execute `?orderly_remote_path` then you'll get documentation for the function, but if it's just confusing then I can drop it

R/migrate.R
```diff
@@ -46,8 +30,13 @@
 ##'   migrated this might come in helpful.
 ##'
 ##' @export
+##' @examples
+##' # Without an orderly repository created by a prevoius version of
```
JamesThompson1729: typo - previous

R/recipe_run.R
```diff
@@ -53,6 +84,15 @@ orderly_run <- function(name, parameters = NULL, envir = NULL,
 
 ##' @export
 ##' @rdname orderly_run
+##' @examples
+##' # The function orderly_data does all the preparation work that
+##' # orderly_run does, but does not run the report but instead
```
JamesThompson1729: _but [...] but [...]_
reads awkwardly
_...but does not run the report, instead it ..._
is maybe better.

R/remote_path.R
```diff
@@ -1,7 +1,25 @@
+##' Create a "handle" for interacting with orderly repositories that
+##' are hosted at a different path.  This might be useful in cases
+##' where you have access to an orderly repository via a network mount
+##' or a syncronised folder (e.g., Dropbox, Box, etc).  More
+##' generally, \code{orderly_remote_path} is implements an interface
```
JamesThompson1729: unnecessary _is_ .

PR #107, opened 2019-06-20
VIMC-1986: improve the db configuration

richfitz:
This PR will harmonise the database configuration with dettl and with the api server configuration by moving the arguments into an explicit block, so that:

```
db:
  driver: pkg::fn
  arg1: value
  arg2: value
```

becomes

```
db:
  driver: pkg::fn
  args:
    arg1: value
    arg2: value
```

This is handled in the same way as the work on database/source, where using the old style is supported for a while, with a warning.

Moving to this layout will allow us to support orderly-level options alongside the drivers in the future (e.g., `cache: false` for preventing caching of database into the `data` directory).

(almost all the lines in the diff here are updating the configurations; there are only a handful of lines of code changed)

### DIFF
--- a/tests/testthat/test-migrate.R
+++ b/tests/testthat/test-migrate.R
@@ -1,7 +1,9 @@
 context("migrations")
 
-
 test_that("0.3.2 -> 0.3.3", {
+  oo <- options(orderly.nowarnings = TRUE)
+  on.exit(options(oo))
+
   path <- unpack_reference("0.3.2")
   cmp <- unpack_reference("0.4.8")
   orderly_migrate(path, to = "0.3.3")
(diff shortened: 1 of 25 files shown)


tests/testthat/test-migrate.R
```diff
@@ -1,7 +1,9 @@
 context("migrations")
 
-
 test_that("0.3.2 -> 0.3.3", {
+  oo <- options(orderly.nowarnings = TRUE)
```
JamesThompson1729: Why are suppressing warnings in these tests?

richfitz: The old style database configuration that is used there (they are reference examples so need to retain them) trigger warnings, but that's expected here

PR #108, opened 2019-06-21
VIMC-2942: Compute all file hashes before run

richfitz:
This PR moves the file hash calculation quite early in the process; it's not at recipe read but I think that's ok because it requires some calculation with directories etc and feels most naturally done when we prepare the new working directory.

This expands the previous "no files changed during run" to cover _all_ files that orderly looks after (somewhat different treatment was required for the dependencies).  The logic with README.md files and how directory listings are computed has been tidied, and I found some undocumented behaviour with globals that I've annotated with  new and existing issues.

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -168,6 +168,12 @@ file_copy <- function(..., overwrite = TRUE) {
 }
 
 
+file_copy_with_path <- function(src, dest) {
+  dir.create(dirname(dest), FALSE, TRUE)
+  file_copy(src, dest)
+}
+
+
 file_move <- function(from, to) {
   ok <- file.rename(from, to)
   if (any(!ok)) {
@@ -698,6 +704,9 @@ file_size <- function(path) {
 
 
 file_info <- function(path, workdir = NULL) {
+  if (is.null(path)) {
+    return(NULL)
+  }
   if (!is.null(workdir)) {
     return(withr::with_dir(workdir, file_info(path)))
   }
--- a/R/util.R
+++ b/R/util.R
@@ -168,12 +168,6 @@ file_copy <- function(..., overwrite = TRUE) {
 }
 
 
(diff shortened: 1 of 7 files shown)


R/util.R
```diff
@@ -138,15 +138,11 @@ dir_create <- function(x) {
 }
 
 hash_files <- function(filenames, named = TRUE) {
-  if (is.null(filenames)) {
```
richfitz: simplified logic here as we do the check for empty file list elsewhere, and this is consistent with the file size check

R/util.R
```diff
@@ -523,20 +519,6 @@ copy_directory <- function(src, as, rollback_on_error = FALSE) {
   }
 }
 
-expand_directory_list <- function(files) {
```
richfitz: there's a subtlety here with the path before and after the source root, so this logic is now in the function where it's used, though it might come back out here once we get global directories working

tests/testthat/test-run.R
```diff
@@ -589,57 +585,55 @@ test_that("orderly_environment", {
   expect_error(orderly_environment(list()), "'envir' must be an environment")
 })
 
-test_that("modify resources", {
-  ## modify 1 resource
+test_that("modify one resource", {
```
richfitz: Logic throughout here is unchanged, just splitting into separate tests

tests/testthat/test-run.R
```diff
@@ -741,3 +735,19 @@ test_that("Can use connections with two databases", {
   p <- orderly_commit(id, root = path)
   expect_true(file.exists(file.path(p, "mygraph.png")))
 })
+
+
+test_that("prevent duplicate filenames", {
```
richfitz: This was never done before, but I'm doing it opportunistically.  Enforcing this will help with eventually deduplicating the archive, and reasoning about where orderly's files come from

PR #109, opened 2019-06-24
Improve error message when pulling dependencies

JamesThompson1729:
When calling `orderly_pull_dependency()` on a report with no dependencies we got a uninformative error.

This PR changes the behaviour to an informative warning  - I didn't think should be an error, since it can't lead to further errors.


R/remote.R
```diff
@@ -40,10 +40,15 @@ orderly_pull_dependencies <- function(name, root = NULL, locate = TRUE,
   path <- file.path(path_src(config$root), name)
   depends <- recipe_read(path, config, FALSE)$depends
 
-  for (i in seq_len(nrow(depends))) {
-    if (!isTRUE(depends$draft[[i]])) {
-      orderly_pull_archive(depends$name[[i]], depends$id[[i]], config,
-                           FALSE, remote)
+  if (is.null(depends)) {
+    orderly_log("warning",
```
richfitz: I'm wary about printing a string that says warning, but does not actually throw one (and similarly, not a big fan of warnings in general because [they're usually informational (use `message`) or really errors (use `stop)](http://blog.obeautifulcode.com/R/A-Warning-About-Warning/).  So, how about, unconditionally

```
orderly_log("depends", msg)
```

with the message printing the number of dependencies that will be checked (`sprintf("%s has %d %s", name, NROW(depends), ngettext(NROW(depends), "dependency", "dependencies"))`)

PR #115, opened 2019-07-30
VIMC-2855: add pulled reports into DB

richfitz:
This PR will add reports into the DB as they are pulled, which is required for commiting any reports that that have imported dependencies.  To make that work we now also recursively download reports


richfitz:
yeah, I think you are right.  I am trying to remember why this is in here - possibly something to do with an idea of "latest"?  Perhaps we can remove it and see what breaks...

PR #117, opened 2019-08-06
Allow README in subdirectories

JamesThompson1729:
At the moment a report structured like this causes problems. If we copy a readme from a subdirectory across as resource there is no way to avoid a problem.
We can't define `meta/README.md` as an artefact or a resource (since it's a readme) so it always appears as an unexpected artefact.
```
src
|___report
  |___orderly.yml
  |___script.R
  |___README.md
  |___meta
    |___dataset_1.csv
    |___dataset_2.csv
    |___README.md # describing the datasets
```

We only should only for check for readmes in the report root directory.
Also the  current code was only checking for exactly **README.md**s and not other variations.


R/recipe_run.R
```diff
@@ -752,7 +752,7 @@ recipe_copy_readme <- function(info, src) {
 
     ## now check if README is an artefact
     artefact_files <- unlist(info$artefacts[, "filenames"], use.names = FALSE)
-    if (any(grepl("README.md", artefact_files, ignore.case = TRUE))) {
+    if (any(grepl("^(?i)readme(|.md)$", artefact_files, ignore.case = TRUE))) {
```
richfitz: what is the `(?i)` doing here?

R/recipe_run.R
```diff
@@ -741,7 +741,7 @@ recipe_copy_readme <- function(info, src) {
 
     ## now check if README is a resource
     if (length(info$resources) > 0) {
-      i <- grepl("README.md", info$resources, ignore.case = FALSE)
```
richfitz: consider above doing

```
  readme_files <- dir(src, pattern = "README(\\.md)?$",
                      ignore.case = TRUE, recursive = TRUE)
```

then adapting the if to be `if (length(readme_files) > 1) {...`

then copying *all* of the readme files over - that would require some variant of `dir_create(file.path(dest, dirname(readme_files))` before the copy and a mapping onto the canonical filename (`sub("README(.\\md)$". "README\\1", readme_files)`) before doing the copy

tests/testthat/test-readme.R
```diff
@@ -102,3 +102,41 @@ test_that("readme db",  {
...
+test_that("allow readme in sub-directory", {
+  path <- prepare_orderly_example("demo")
+  ## in report directory create a file called README.md
+  report_path <- file.path(path, "src", "use_resource")
+  # add a readme fiel to the meta data directory
+  file.create(file.path(report_path, "meta", "README.md"))
+
+  #rewrite yml to include extra readme file
+  yml_path <- file.path(report_path, "orderly.yml")
+  yml <- c("data:",
```
richfitz: if you make the change above then all the readmes will come along for free

R/recipe_run.R
JamesThompson1729: my understanding of regex is that this should make it case insensitive so we get README and readme.

richfitz: ok that's new to me - could you use the `ignore.case = TRUE` argument instead to be slightly less cryptic?

R/recipe_run.R
JamesThompson1729: Should the logic be '*copy across all READMEs we find*'?

This came up because there was report that listed a directory as a resource that contained a readme and there was no way to get rid of the unexpected artefact warning.

Do we want to support the use case where a report directory contains a file that is not copied across?

JamesThompson1729:
Changes look good, happy to merge this in

PR #118, opened 2019-08-08
strip out trailing slashes from resource directories

JamesThompson1729:
If a report has a resource directory specified by
```
resources:
  - data/
```

The report will run correctly with all the contents of data copied correctly.
The files in data will be listed as unexpected artefacts since orderly the files are copied to
```
data/filename.csv
```
but orderly expects the file to be
```
data//filename.csv
```
[This might also cause problems running this report on a windows OS]

This fix removes a trailing slash - we could probably go further sanitizing filenames (and other entries) in the yaml

### DIFF
--- a/R/recipe_read.R
+++ b/R/recipe_read.R
@@ -217,10 +217,6 @@ recipe_read_check_resources <- function(x, filename, path) {
   trailing <- grepl(pattern = "(\\/)$", x)
   bad_resource <- is_dir & trailing
   if (any(bad_resource)) {
-    warning_msg <- sprintf("Resource %s a traling slash:%s",
-                   ngettext(sum(bad_resource), "directory has",
-                                               "directories have"),
-                   paste(squote(x[bad_resource]), collapse = ", "))
     x[bad_resource] <- sub("(\\/)$", "", x[bad_resource])
   }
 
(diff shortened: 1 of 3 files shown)


R/recipe_read.R
```diff
@@ -210,6 +210,21 @@ recipe_read_check_resources <- function(x, filename, path) {
   if (is.null(x)) {
     return(NULL)
   }
+
+  # make sure that a directory resource does not have a trailing /
+  # There is much more sanitation that could be done here...
+  is_dir <- is_directory(file.path(path, x))
```
richfitz: I think either this should be ok (no warning, strip the slash and continue) or not ok (throw an error) but not a warning - either we want the user to do something about it or not

richfitz:
(Also the version number should be incremented, up to you if you think this warrants a news item)

R/recipe_read.R
JamesThompson1729: Warning removed - We might want to add more file name sanitation; no one is going to read dozens of warnings

R/recipe_read.R
```diff
@@ -210,6 +210,20 @@ recipe_read_check_resources <- function(x, filename, path) {
...
+  trailing <- grepl(pattern = "(\\/)$", x)
+  bad_resource <- is_dir & trailing
+  if (any(bad_resource)) {
+    warning_msg <- sprintf("Resource %s a traling slash:%s",
```
richfitz: I think 220-223 can come out here

PR #119, opened 2019-08-29
VIMC-2953: orderly takes responsibility for copying directory in pull_archive

richfitz:
This PR will change the behaviour of the pull_archive function and so needs to be coordinated with an orderlyweb PR https://github.com/vimc/orderlyweb/pull/3.

The motivation here is that copying directories is hard, but it's easy to just point at one. So now the remote driver never modifies the destination orderly repo (which seems safer) and instead tells orderly where to find the report


R/remote.R
```diff
@@ -101,27 +101,24 @@ orderly_pull_archive <- function(name, id = "latest", root = NULL,
...
+    ## columns) then we'll need to deal with that when checking
+    ## dependencies too.
+    orderly_pull_resolve_dependencies(path, remote, config)
 
+    ## Only migrate after dependencies have been resolved because
+    ## some migrations do check dependencies.
+    migrate_single(path, config)
+
+    withCallingHandlers({
+      copy_directory(path, dest)
```
JamesThompson1729: How bad is the situation where the migrate succeeds but the copy_directory fails?
We end up with a report in the database but not in the archive.

I suppose if copy_directory fails there are bigger problems...

richfitz: I think the only way that this failure likely happens is that we're out of disk space (as we've already found that we can read from the source directory and that it contains a reasonable orderly report). Recovering would be a case of rebuilding the db.  This is hopefully uncommon enough that we can ignore it

PR #124, opened 2019-09-16
VIMC-731: Implement deduplication

richfitz:
This PR will provide the first tools that support deduplicating an orderly archive.

There is more to add here eventually - in particular for deduplicating archives where the creators have modified file inputs (we should skip these gracefully).

My plan is that once this is working we'll run it on science/uat a bit and try and establish that things are indeed ok before doing it in production


R/deduplicate.R
```diff
@@ -0,0 +1,206 @@
...
+  assert_is(x, "orderly_deduplicate_info")
+
+  ## This is the only safe thing to do here:
+  if (length(unique(x$files$device_id)) > 1L) {
+    stop("Can't deduplicate as your orderly archive spans multiple devices")
+  }
+
+  ## Likely too strict as we can just skip these
+  if (!all(x$files$unchanged)) {
+    stop("Can't deduplicate files that have been modified")
```
JamesThompson1729: Should we print out which files have been modified?
(I got this error when I ran deduplicate on my machine and I don't know which files have been modified.)

R/deduplicate.R
```diff
@@ -0,0 +1,206 @@
...
+##' repeatedly used in different reports, or outputs that end up being
+##' the same in multiple reports.  If you run the deduplication with
+##' \code{dry_run = TRUE}, an indication of the savings will be
+##' printed.
+##'
+##' @title Deduplicate an orderly archive
+##'
+##' @inheritParams orderly_list
+##'
+##' @param dry_run Logical, indicating if the deduplication
```
JamesThompson1729: missing word here.

R/deduplicate.R
```diff
@@ -0,0 +1,206 @@
...
+##' immediately reflected in the other, but there is nothing to
+##' indicate that the files are linked!
+##'
+##' This approach is worth exploring if you have large files that are
+##' outputs of one report and inputs to another, or large inputs
+##' repeatedly used in different reports, or outputs that end up being
+##' the same in multiple reports.  If you run the deduplication with
+##' \code{dry_run = TRUE}, an indication of the savings will be
+##' printed.
+##'
```
weshinsley: General OS comment - For the sake of getting orderly on to CRAN, how important is it that orderly should be possible to run on Windows? It sounds like this would be a significant (and awkward) programming investment, and probably one for me to tackle if that were a desired thing.

R/deduplicate.R
```diff
@@ -0,0 +1,206 @@
+##' Deduplicate an orderly archive.  Deduplicating an orderly archive
+##' will replace all files that have the same content with "hard
+##' links".  This is reliable only on Unix-like systems.  If you use
+##' this feature, it is \emph{very important} that you treat your
```
weshinsley: Can you clarify "reliable only on Unix-like systems" ... does this mean we shouldn't do it, or can't do it, or.... 

I'm not super-familiar with this, but Windows has apparently supported hard links (of some sort) since Win 7 (https://www.howtogeek.com/howto/16226/complete-guide-to-symbolic-links-symlinks-on-windows-or-linux/) - so I would have thought Windows solution should be doable at some stage.... if we wanted to.

R/deduplicate.R
richfitz: I've added a list here - it might get truncated for very large lists of files. Let me know how that goes on your machine

R/deduplicate.R
richfitz: There is probably nothing stopping this running on windows, and if you wanted to check it please do.  The issue can you see how this gets on with deduplicating on a win7/win10 machine by removing the check. The things to watch out for are:

* how does the inode check go?
* does the hardlink work?
* can we tell that the hardlink worked?

R/deduplicate.R
```diff
@@ -0,0 +1,206 @@
...
+##' orderly archive as read-only (though you should be anyway) as
+##' changing one copy of a linked file changes all the other instances
+##' of it - the files are literally the same file.
+##'
+##' This function will alter your orderly archive.  Ordinarily this is
+##' not something that should be done, so we try to be careful.  In
+##' order for this to work, it is \emph{very important} to treat your
+##' orderly archive as read-only generally.  If your canonical orderly
+##' archive is behind OrderlyWeb this will almost certainly be the
```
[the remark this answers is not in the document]
richfitz: these are split in the rendered docs by the parameters and usage sections so I wanted to repeat it :)

R/deduplicate.R
richfitz: means currently disabled, not sure how well it would work (see other comments for details)

R/deduplicate.R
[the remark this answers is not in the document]
weshinsley: Ah yes...!

PR #125, opened 2019-09-17
Fix windows zip issue, and skip some non-critical win-failing tests

weshinsley:


R/testing.R
```diff
@@ -206,6 +206,19 @@ build_git_demo <- function() {
 
   git_checkout_branch("master", root = path)
 
+  # On windows, make ".git" folder non-hidden,
```
richfitz: It might be nice to isolate the ick a bit by putting it into it's own function - it would just take `path` as an argument

weshinsley: Try now...

PR #126, opened 2019-09-20
VIMC-3137: Remove uses of pandoc

richfitz:
While `rmarkdown::render` is certainly a nicer way of doing this, it induces an error in the pandoc system package for tests, which will be a problem on CRAN (I believe it does not exist on Solaris, for example).

This PR uses plain knitr for the translation

### DIFF
--- a/inst/examples/demo/src/html/myreport.Rmd
+++ b/inst/examples/demo/src/html/myreport.Rmd
@@ -1,10 +1,3 @@
----
-title: "An example report"
-author: "orderly"
-date: "20 February 2018"
-output: html_document
----
-
 ## Introduction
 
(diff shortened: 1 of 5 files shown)


JamesThompson1729:
Should the tests suppress the knitr comments?
It's not a big deal but `devtools::test()` outputs:
```
✔ |  17       | cleanup [1.1 s]
✔ |  41       | config [0.1 s]
  |......................                                           |  33%
  ordinary text without R code

  |...........................................                      |  67%
label: unnamed-chunk-1
  |.................................................................| 100%
  ordinary text without R code


✔ |  33       | db [5.4 s]
✔ |   6       | deprecated functions [0.8 s]
✔ |   8       | envir
```
which is a bit annoying.

richfitz:
thanks, will look at that

richfitz:
tests should be quieter now, thanks

PR #130, opened 2019-10-01
VIMC-2961: Allow resources in subdirectories (and with different names)

richfitz:
This PR will allow global resources to be organised into subdirectories.  Previously this was not really possible as it was ambiguous what

```
global_resources:
  - path/file.csv
```

should do - should it copy to the current report as `path/file.csv` or as `file.csv` (this ambiguity does not exist for ordinary resources as this is just a declaration that they exist).

To solve this we take the same approach as dependencies where files can be renamed:

```
global_resources:
  file.csv: path/file.csv
```

The lhs is the destination - that is the order used in the dependencies.  This approach allows arbitrary renaming, which is quite nice.

There are changes to the schema as a result of this - a new table exists to map the original names.

I do not believe that any report currently uses global resources so this should have minimal effect at the moment.  A migration is included to be on the safe side though.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: orderly
 Title: Lightweight Reproducible Reporting
-Version: 0.7.14
+Version: 0.7.15
 Description: Order, create and store reports from R.  By defining a
     lightweight interface around the inputs and outputs of an
     analysis, a lot of the repetitive work for reproducible research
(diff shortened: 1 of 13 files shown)


JamesThompson1729:
I've updated the vignette to reflect these changes.
And there is vimc branded report template in montagu reports that I will update (It has never been used). It still work but produces the deprecation warning

richfitz:
See https://github.com/vimc/montagu-reports/pull/342 / VIMC-3151 for the updfates to montagu-reports

PR #131, opened 2019-10-03
VIMC-3157: support cleaning up old migration files

richfitz:
This PR will add an option to `orderly_migrate` (and the cli) to clean up old migration backup files (like `orderly_run_0.6.0.rds`). These clutter up the downloaded zips and should be safe to delete eventually

### DIFF
--- a/R/migrate.R
+++ b/R/migrate.R
@@ -29,6 +29,11 @@
 ##'   contains unusual copies of orderly archives that can't be
 ##'   migrated this might come in helpful.
 ##'
+##' @param clean Logical, where \code{TRUE} (and where the migration
+##'   was successful and \code{dry_run} is \code{FALSE}) orderly will
+##'   clean up all migration backup files.  Use this periodically to
+##'   clean up the archive.
+##'
 ##' @export
 ##' @examples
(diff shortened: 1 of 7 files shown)


R/migrate.R
```diff
@@ -246,3 +255,17 @@ migrate_single <- function(path, config) {
...
+migrate_clean <- function(config, dry_run) {
+  files <- list.files(file.path(config$root, "archive"),
+                      "^orderly_run_([0-9]+\\.){3}rds",
+                      recursive = TRUE, full.names = TRUE)
+  size <- pretty_bytes(sum(file_size(files)))
+  orderly_log("clean",
+              sprintf("%d backup files to delete (%s)", length(files), size))
+  if (!dry_run) {
+    file.remove(files)
+  }
```
JamesThompson1729: Should we confirm that the files have been deleted?
Currently we get 
```
orderly::orderly_migrate(clean=TRUE)
[ clean      ]  1216 backup files to delete (25.75 MB)
```
But orderly does not tell me that the files actually have been deleted.

richfitz: yes, good point!

PR #133, opened 2019-10-04
Quicker git_info on CRAN

weshinsley:


tests/testthat/test-util.R
```diff
@@ -161,6 +161,15 @@ test_that("git", {
   info <- git_info(path)
   expect_true(all(c("sha_short", "sha", "branch", "status") %in% names(info)))
 
+  nogit <- (Sys.getenv("NOT_CRAN") != "true")
```
richfitz: This should be done with withr, and always test against a valud path argument.

```
withr::with_options(
  list(orderly.nogit = TRUE),
  expect_null(git_info(path)))
```

you should do this in a new test block and start by unpacking the orderly git demo

tests/testthat/helper-orderly.R
```diff
@@ -130,3 +130,7 @@ new_counter <- function() {
     e$x
   }
 }
+
+if (Sys.getenv("NOT_CRAN") != "true") {
+  options(orderly.nogit = TRUE)
+}
```
richfitz: could you fix this newline issue please?

tests/testthat/test-util.R
weshinsley: Done!

tests/testthat/helper-orderly.R
weshinsley: Fixed

tests/testthat/test-util.R
```diff
@@ -173,6 +173,19 @@ test_that("git", {
                info[c("sha_short", "sha", "branch")])
 })
 
+test_that("git_info non-cran", {
+  path <- unzip_git_demo()
```
richfitz: sorry, looks like this needs a `skip_on_cran()` as the `unzip_git_demo` will otherwise not work. Should also add a `skip_if_no_git` in here too

PR #140, opened 2019-10-08
vimc-2174

JamesThompson1729:


DESCRIPTION
```diff
@@ -35,6 +35,7 @@ Imports:
     yaml,
     zip (>= 2.0.0)
 Suggests:
+    crayon,
```
richfitz: This is being used unconditionally so needs to be in Imports

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
+##' @title Get the dependencies for a given report from the database
+##'
+##' @param name the name of the report
+##' @param id the id of the report, if omitted, use the id of the latest report
+##' @param upstream A boolean indicating if we want to move up or down the tree
+##' @param con A connection to a database
+get_dependencies_db <- function(name, id, upstream, con, list_all = FALSE) {
```
richfitz: All exported orderly functions have `orderly_` as their name, and in general I do not like the use of `get` here. How about simply `orderly_dependencies` (or perhaps `orderly_report_dependencies`)

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
+##' @title Get the dependencies for a given report from the database
+##'
+##' @param name the name of the report
+##' @param id the id of the report, if omitted, use the id of the latest report
+##' @param upstream A boolean indicating if we want to move up or down the tree
```
richfitz: is there a case where we might want to do *both* directions? In that case, even if not implemented now we'd be better off with a character argument here (`direction = "upstream"` or `direction = "downstream"`)

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+##' @param con A connection to a database
+get_dependencies_db <- function(name, id, upstream, con, list_all = FALSE) {
+  ## now construct the SQL query
+  if (upstream) {
+    filt_qry <- sprintf("depends.report_version='%s'", id)
```
richfitz: please avoid excessive abbreviation (filt, qry, etc)

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+  } else {
+    filt_qry <- sprintf("report_version.id='%s'", id)
+  }
+
+  sql_qry <- paste(c("SELECT",
```
richfitz: Can we move this into the sql directory in inst? If not, there seem to be more commas here than needed

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+  db_ret <- DBI::dbGetQuery(con, sql_qry)
+
+  if (nrow(db_ret) == 0) {
+    return(NULL)
+  }
+
+  db_ret$is_latest <- sapply(db_ret$report_version,
+                               is_latest_in_db, con = con)
+  if (!list_all) {
+    db_ret <- db_ret[which(db_ret$is_latest), ]
```
richfitz: no need for `which` here

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+               "report_version_artefact.report_version=report_version.id)",
+               "WHERE", filt_qry), collapse = " ")
+
+  db_ret <- DBI::dbGetQuery(con, sql_qry)
+
+  if (nrow(db_ret) == 0) {
+    return(NULL)
+  }
+
+  db_ret$is_latest <- sapply(db_ret$report_version,
```
richfitz: please avoid sapply - here use `vlapply`

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+                               is_latest_in_db, con = con)
+  if (!list_all) {
+    db_ret <- db_ret[which(db_ret$is_latest), ]
+  }
+
+  # if we're going uptree
+  if (upstream) {
+    return(unique(db_ret$id))
```
richfitz: please do not use explicit return statements if it is not an early exit (here and elsewhere)

R/dependencies.R
richfitz: oh, I see this is not actually an exported function - in which case I think these roxygen blocks all require `@noRd` or similar to prevent creation of `man` files

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+  } else {
+    return(unique(db_ret$report_version))
+  }
+}
+
+##' @title Get the id of the latest version of a report
+##'
+##' @param name the name of the report
+##' @param con A connection to a database
+get_latest_by_name <- function(con, name) {
```
richfitz: The `get` here is not helpful - something like `report_latest_version` might capture intent better.  However we have similar functions in query.R already - could you harmonise?

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+  sql_qry <- paste("SELECT", "id, report, date FROM report_version",
+                   "WHERE", sprintf("report='%s'", name),
+                   "AND", "date=(SELECT MAX(date)",
+                   "FROM", "report_version",
+                   sprintf("WHERE report='%s')", name))
+  db_ret <- DBI::dbGetQuery(con, sql_qry)
+  return(db_ret)
+}
+
+get_latest_by_id <- function(con, id) {
```
richfitz: this could be harmonised with code in (query.R & nearby functions)

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+  }
+}
+
+##' @title Is the id the latest version of the report in the database
+##'
+##' @param id the id of the report
+##' @param con A connection to a database
+##'
+##' @return A boolean TRUE if the report is the latest version
+is_latest_in_db <- function(con, id) {
```
richfitz: do these all warrant being free functions?

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+
+  db_chd <- DBI::dbGetQuery(con, sql_chd_qry)
+
+  ## this reports uses no artefacts - so can never be out of date
+  if (nrow(db_chd) == 0) {
+    return(FALSE)
+  }
+
+  ## we iterate over rows of the data frame and make sure the artefact match
+  for (i in 1:nrow(db_chd)) {
```
richfitz: please do not use `1:n` for sequences - use `seq_len`

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+    filename <- db_chd$filename[i]
+    file_hash <- db_chd$file_hash[i]
+    report_id <- db_chd$report_version[i]
+    ## we need to find the id latest version of the report with id = report_id
+    if (!db_chd$is_pinned[i]) { ## if not pinned...
+      latest_id <- get_latest_by_id(con, report_id) ## ..use latest
+    } else { ## if pinned to..
+      if (!db_chd$is_latest[i]) { ## ..not latest..
```
richfitz: this if/else has the same body in both branches

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+             "FROM", "(report_version",
+             "INNER JOIN", "report_version_artefact", "ON",
+                           "report_version.id = report_version_artefact.report_version",
+             "INNER JOIN", "file_artefact", "ON",
+                           "file_artefact.artefact = report_version_artefact.id)",
+             "WHERE", par_qry), collapse = " ")
+
+    db_par <- DBI::dbGetQuery(con, sql_par_qry)
+    i <- which(db_par$filename == filename)
+    if (db_par$file_hash[i] != file_hash)
```
richfitz: please always use braces, even for one-line statements. The `which` in the line above appears unneeded

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+
+##' @title Recursively check that none of the parents share the same name as the
+##' current report.
+##'
+##' @param parent_vertex The vertex whose name we want to check
+##' @param name The name we want to match
+##'
+##' @return A boolean TRUE if there is a match (bad!) FALSE if the is no match
+##' (good!)
+check_parents <- function(parent_vertex, name) {
```
richfitz: Can you elaborate on when this is `FALSE`?

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+##' @param locate
+##' @param upstream A boolean indicating if we want to move up or down the tree
+##'
+##' @return An R6 tree object
+build_tree <- function(name, id, depth = 100, parent = NULL,
+                       tree = NULL, con, upstream = FALSE, list_all = FALSE) {
+  ## this should never get triggered - it only exists the prevent an infinite
+  ## recursion
+  if (depth < 0) {
+    stop("The tree is very large or degenerate.")
```
richfitz: This is not a helpful message for users - what leads to this situation? Why is large a problem?

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+  }
+
+  if (!is.null(parent)) {
+    circ <- check_parents(parent, name)
+    if (circ) {
+      tree$message <- "There appears to be a circular dependency."
```
richfitz: is this an error condition?

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+      return(tree)
+    }
+  }
+
+  # make sure a report with this name exists
+  db_reports <- DBI::dbGetQuery(con, "SELECT name FROM report")
+  if (!(name %in% db_reports$name)) {
+    stop("This report does not exist")
```
richfitz: Can this happen? If we're working purely with the DB then the foreign keys are doing all the referential integrity work for us.

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+  }
+
+  # do we need to find the latest version of the report?
+  if (id == "latest") {
+    id <- get_latest_by_name(con, name)$id
```
richfitz: everywhere you use `get_latest_by_name` you are extracting `$id` so probably refactor the function to just return that

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+  } else {
+    db_id <- id_to_name(con, id)
+    if (is.null(db_id)) {
+      stop(sprintf("No report with id %s in the database", id))
+    } else {
```
richfitz: collapse the `else` and the `if`

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+      if (name != id_to_name(con, id)) {
+        stop(sprintf("id %s does not match report name %s", id, name))
+      }
+    }
+  }
+
+  if (upstream) {
```
richfitz: both branches have the same content?

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+      str <- paste(str, crayon::red(self$message), "\n", collapse = "")
+      self$message <- NULL
+    }
+
+    console_colour <- if (vertex$out_of_date) {crayon::red} else {crayon::blue}
+
+    if (length(fvector) == 0) {
+      str <- paste(str, console_colour(sprintf("%s", vertex$format())), "\n", collapse = "")
+    } else {
+      spacing <- paste(ifelse(head(fvector, -1), "| ", "  "), collapse = "")
```
richfitz: do not use ifelse

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+  },
+  # this is stupidly hacky to get the formatting right
+  format_helper = function(vertex = self$root, fvector = c(), str="") {
+    ## if the tree has a warning message, append it to the front in the red
+    if (!is.null(self$message)) {
+      str <- paste(str, crayon::red(self$message), "\n", collapse = "")
+      self$message <- NULL
+    }
+
+    console_colour <- if (vertex$out_of_date) {crayon::red} else {crayon::blue}
```
richfitz: no braces on a one-line if/else

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+Tree <- R6::R6Class("Tree", list(
+  vertices = NULL,
+  root = NULL,
+  message = NULL,
+  initialize = function(root) {
+    self$vertices <- append(self$vertices, list(root))
+    self$root <- root
+    self$message <- NULL
+  },
+  add_child = function(parent, name, id, out_of_date) {
```
richfitz: one blank line between all methods

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+      if (vertex$out_of_date) {
+        child$out_of_date = TRUE
+      }
+    }
+
+    propagate(child, upstream)
+
+    if (upstream) {
+      if (child$out_of_date) {
+        vertex$out_of_date = TRUE
```
richfitz: no equals for assignment

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+##' @param name the name of the report
+##' @param id the id of the report, if omitted, use the id of the latest report
+##' @param con A connection to a database
+##' @param root
+##' @param locate
+##' @param upstream A boolean indicating if we want to move up or down the tree
+##' @param propagate A boolean indicating if we want to propogate out of date
+##'                  through the tree
+##'
+##' @export
```
richfitz: This has not yet been exported - you need to re-document the package

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+orderly_build_dep_tree <- function(name, id = "latest", root = NULL,
```
richfitz: `orderly_dependency_tree` perhaps - avoids the abbreviation and the `build` part is not needed

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+##'
+##' @param name the name of the report
+##' @param id the id of the report, if omitted, use the id of the latest report
+##' @param con A connection to a database
+##' @param root
+##' @param locate
+##' @param upstream A boolean indicating if we want to move up or down the tree
+##' @param propagate A boolean indicating if we want to propogate out of date
+##'                  through the tree
+##'
```
richfitz: please add an example - you can use the demo orderly example

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+    for (i in 1:length(vertex$children)) {
+      child <- vertex$children[[i]]
+
+      reports <- out_ot_date_reports(child, reports)
+    }
+  }
+  reports
+}
+
+##' @title Print the dependency tree for a given report using orderly log
```
richfitz: Please move this to the top of the file

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+##'
+##' @param name the name of the report
+##' @param id the id of the report, if omitted, use the id of the latest report
+##' @param con A connection to a database
+##' @param root
```
richfitz: Use @inheritParams to get these added - see `orderly_run` for example

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+      child <- vertex$children[[i]]
+
+      reports <- out_ot_date_reports(child, reports)
+    }
+  }
+  reports
+}
+
+##' @title Print the dependency tree for a given report using orderly log
+##'
```
richfitz: Please expand the discussion of what this function is for, how one uses it, etc

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+    } else {
+      spacing <- paste(ifelse(head(fvector, -1), "| ", "  "), collapse = "")
+      str <- paste(str, console_colour(sprintf("%s|___%s", spacing, vertex$format())), "\n", collapse = "")
+    }
+
+    if (length(vertex$children) > 0) {
+      for (i in 1:length(vertex$children)) {
+        child <- vertex$children[[i]]
+
+        if (i != length(vertex$children)) {
```
richfitz: this mgiht be simpler as

```
thing <- i != length(vertex$children)
str <- self$format_helper(child, c(fvector, thing), str)
```

though I am unclear what the second argument here represents at this point

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+    self$message <- NULL
+  },
+  add_child = function(parent, name, id, out_of_date) {
+    child <- Vertex$new(parent, name, id, out_of_date)
+    parent$add_child(child)
+    self$vertices <- append(self$vertices, list(child))
+    return(child)
+  },
+  # this is stupidly hacky to get the formatting right
+  format_helper = function(vertex = self$root, fvector = c(), str="") {
```
richfitz: this should be a private method, and we should think about making it not stupidly hacky, or at least explaining what the problem is

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+
+    if (upstream) {
+      if (child$out_of_date) {
+        vertex$out_of_date = TRUE
+      }
+    }
+  }
+}
+
+Vertex <- R6::R6Class("Vertex", list(
```
richfitz: this and tree seem totally unrelated to everything else in the file, which is good. Can you move them to their own file and add some testing to them separate to the rest of the dependency code

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+  for (child in vertex$children) {
+    if (!upstream) {
+      if (vertex$out_of_date) {
+        child$out_of_date = TRUE
+      }
+    }
+
+    propagate(child, upstream)
+
+    if (upstream) {
```
richfitz:
```
if (upstream && child$out_of_date) {
  vertex$out_of_date <- TRUE
}
```

or even

```
vertex$out_of_date <- upstream && child$out_of_date
```

and similar above. there's quite a bit of noise that can be removed like this

R/dependencies.R
```diff
@@ -0,0 +1,410 @@
...
+##' @param upstream A boolean indicating if we want to move up or down the tree
+##' @param propagate A boolean indicating if we want to propogate out of date
+##'                  through the tree
+##'
+##' @export
+orderly_build_dep_tree <- function(name, id = "latest", root = NULL,
+                                   locate = TRUE, upstream = FALSE, con = NULL,
+                                   propagate = FALSE, max_depth = 100,
+                                   list_all = FALSE) {
+  if (is.null(con)) {
```
richfitz: I don't see a use case for accepting both root/locate and con - drop the con argument and make the connection unconditional

JamesThompson1729:
I have made all the changes.
We use the the local report database to find dependencies, this causes a few problems 
 * If a report has been rewritten to change the dependencies this wont get picked up since we never read the yaml files.
 * If a report has been re-run nothing will depend on it. To get around this I added a _"previous"_ flag to _"id"_ which gets the id of the second most recent version of a report.

The two basic use cases are:
 * We want to update a report and need to know which _'upstream'_ reports it depends on - 
`print(orderly_build_dep_tree("report_name", direction = "upstream"))`
This will print the dependency tree with out-of-date reports marked in red.

 * We have re-run and report and we want to re-run everything that depends on it -
`print(orderly_build_dep_tree("report_name", direction = "downstream", id = "previous"))`
we have to use the _previous_ because once a report has been re-run nothing depends on yet. So we check what depended on the previous version of the report.

R/dependencies.R
richfitz: Still pending

R/dependencies.R
richfitz: Still pending

R/dependencies.R
richfitz: still pending

R/dependencies.R
richfitz: Still pending

R/dependencies.R
JamesThompson1729: This was to protect against a scenario where the tree somehow has circular dependency.
I could not guarentee that this impossible. and I wanted to avoid a situation where we get into an infinite recursion where orderly hangs and the user has no idea what went wrong.

R/dependencies.R
JamesThompson1729: I think it can only be hit if the user passes in an invalid name. I've moved this logic to *orderly_build_dep_tree*, so we only check once when before the recursive steps begin.

R/dependencies.R
```diff
@@ -0,0 +1,471 @@
...
+##'    on it.
+##' 2. We have a report that we want to re-run to ensure uses the latest
+##'    information. Which other reports are used (directly or indirectly) by
+##'    this report?
+##'
+##' This function display this information in an easily readable format.
+##' Allowing users to see the dependency tree and which reports are out of date 
+##' and need to be re-run.
+##'
+##' Remark
```
richfitz:
```
##' @section Remark:
```

R/dependencies.R
```diff
@@ -0,0 +1,471 @@
...
+##' @title Given a tree return a list of reports to be re-run (and the order
+##' that they should be re-run)
+##'
+##' @param tree A dpendency tree object from orderly_dependency_tree
+##'
+##' @return a list of report names to be re-run. First report to rerun first
+##' @export
+orderly_out_of_date_reports <- function(tree) {
+  types <- class(tree)
+  if (!("Tree" %in% types)) {
```
richfitz: `assert_is(tree, "Tree")`

R/dependencies.R
```diff
@@ -0,0 +1,471 @@
...
+
+  sql_query <- paste(c(sql_query_file, "WHERE", filter_query), collapse = " ")
+
+  query_return <- DBI::dbGetQuery(con, sql_query)
+
+  if (nrow(query_return) == 0) {
+    return(NULL)
+  }
+
+  query_return$is_latest <- vapply(query_return$report_version,
```
richfitz: vlapply

R/dependencies.R
```diff
@@ -0,0 +1,471 @@
...
+##' @return a one line dataframe with columns:
+##'         id - the report id
+##'         report - the name of the report (matches the argument name)
+##'         date - the date the report was run
+##'
+##' @noRd
+get_latest_by_name <- function(con, name) {
+  reports <- get_ids_by_name(con, name)
+
+  (reports$id)[1]
```
richfitz: no need for parens here

R/dependencies.R
```diff
@@ -0,0 +1,471 @@
...
+##'
+##' @noRd
+get_previous_by_name <- function(con, name) {
+  reports <- get_ids_by_name(con, name)
+
+  if (nrow(reports) < 2) {
+    stop(sprintf("There is only one version of %s", name))
+  }
+
+  (reports$id)[2]
```
richfitz: No parens

PR #149, opened 2019-11-20
vimc-3251: Clean path in orderly_test_start message

richfitz:
This PR creates instructions in `orderly_test_start` that are copy/pasteable


richfitz:
There's an image in the [YouTrack issue](https://vimc.myjetbrains.com/youtrack/issue/VIMC-3251) showing the problem. Pasting, for example,

```
setwd("C:\Users\me\Documents\Projects\reports/draft/my-report")
```

will fail because R interprets `\U` as a control sequence (like `\n` or `\r`) and not as a literal backslash-then-U. Python has ["raw strings"](https://www.journaldev.com/23598/python-raw-string) for this, but they're not in R unfortunately.  So we must either double the slashes or turn them into forward slashes.

PR #150, opened 2019-12-17
vimc-3302: more flexible database selection

richfitz:
This PR aims to address the database instance problem discussed in more detail in the corresponding ticket https://mrc-ide.myjetbrains.com/youtrack/issue/VIMC-3302

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# orderly 1.0.6
+
+* Database configurations now support the concept of "instances" to allow switching between different versions of a database (e.g., production and staging) without manually altering the configuration or environment variables. Functions `orderly::orderly_run`, `orderly::orderly_db`, `orderly::orderly_test_start` and `orderly::orderly_data` all get a `instance` argument to support this (VIMC-3302).
+
 # orderly 1.0.5
 
 * The metadata now includes the state of `.Random.seed`, if present (VIMC-3375)
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # orderly 1.0.6
 
-* Database configurations now support the concept of "instances" to allow switching between different versions of a database (e.g., production and staging) without manually altering the configuration or environment variables. Functions `orderly::orderly_run`, `orderly::orderly_db`, `orderly::orderly_test_start` and `orderly::orderly_data` all get a `instance` argument to support this (VIMC-3302).
+* Database configurations now support the concept of "instances" to allow switching between different versions of a database (e.g., production and staging) without manually altering the configuration or environment variables. Functions `orderly::orderly_run`, `orderly::orderly_db`, `orderly::orderly_test_start` and `orderly::orderly_data` all get an `instance` argument to support this (VIMC-3302).
 
 # orderly 1.0.5
 
--- a/R/config.R
+++ b/R/config.R
@@ -231,9 +231,6 @@ config_read_db <- function(name, info, filename) {
     check_fields(dat, label, "driver", optional)
 
     if (!is.null(dat$instances)) {
-      if (identical(name, "destination")) {
-        stop("Instances only supported for 'database', not 'destination'")
-      }
       assert_named(dat$instances, TRUE, paste0(label, ":instances"))
       for (i in names(dat$instances)) {
         assert_named(dat$instances[[i]], TRUE, paste0(label, ":instances:", i))
--- a/R/config.R
+++ b/R/config.R
@@ -21,31 +21,6 @@ orderly_config_read_yaml <- function(filename, root) {
   ## directories; things like the sqlite path.  And I want to be able
   ## to encrypt and decrypt things like passwords.  So we probably
   ## cannot be totally opaque when reading information in.
-
-  driver_config <- function(name) {
-    if (identical(name, "destination") && is.null(info[[name]])) {
-      dat <- list(driver = "RSQLite::SQLite",
-                  args = list(dbname = "orderly.sqlite"))
-    } else {
-      dat <- info[[name]]
-    }
-    label <- sprintf("%s:%s:driver", filename, paste(name, collapse = ":"))
-    driver <- check_symbol_from_str(dat$driver, label)
-    if ("args" %in% names(dat)) {
-      args <- dat$args
-    } else {
-      if (!info$database_old_style) {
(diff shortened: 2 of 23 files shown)


R/config.R
```diff
@@ -21,31 +21,6 @@ orderly_config_read_yaml <- function(filename, root) {
   ## directories; things like the sqlite path.  And I want to be able
   ## to encrypt and decrypt things like passwords.  So we probably
   ## cannot be totally opaque when reading information in.
-
-  driver_config <- function(name) {
```
richfitz: this has been lifted out into its own function, below

R/db.R
```diff
@@ -27,6 +27,16 @@
 ##'   be validated on open (currently only applicable with \code{type
 ##'   = "destination"}).  This is primarily intended for internal use.
 ##'
+##' @param instance Used only by \code{type = "source"}, and used to
```
richfitz: please check to make sure this makes sense.

vignettes/orderly.Rmd
```diff
@@ -571,6 +571,38 @@ would look up the field `password` at the path
 `orderly_config.yml`, in the contents of an environment variable or
 in `orderly_envir.yml` (currently this only uses the vault [version 1 key-value storage](https://www.vaultproject.io/docs/secrets/kv/kv-v1.html))
 
+The above approach can be used to switch databases by using different environmental variables, but that can become tiresome.  If you have multiple database "instances" corresponding to different realisations of the same logical database (e.g., production and staging), then you can configure and switch between these by using the `instances` field.  List commmon arguments within the `args` block as before, then add named instances:
```
richfitz: please check to make sure the approach is explained clearly enough

richfitz:
Support added to montagu-reports in https://github.com/vimc/montagu-reports/compare/vimc-3318 and https://github.com/vimc/montagu-orderly-web/pull/8

NEWS.md
```diff
@@ -1,3 +1,7 @@
+# orderly 1.0.4
+
+* Database configurations now support the concept of "instances" to allow switching between different versions of a database (e.g., production and staging) without manually altering the configuration or environment variables. Functions `orderly::orderly_run`, `orderly::orderly_db`, `orderly::orderly_test_start` and `orderly::orderly_data` all get a `instance` argument to support this (VIMC-3302).
+
```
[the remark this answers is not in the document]
weshinsley: And this is probably #orderly 1.0.5 now

vignettes/orderly.Rmd
weshinsley: Looks fine to me,.

R/recipe_run.R
```diff
@@ -48,7 +48,18 @@
...
+##' @param instance Select instance of the source database to be used,
+##'   where multiple instances are configured.  Several options here
+##'   are available - use a single \emph{unnamed} character string to
+##'   indicate an instance to match.  If given, then this name must be
+##'   present in all databases that support instances, and will be
+##'   ignored by all that do not.  Otherwise, provide \emph{named}
+##'   character vector with the name being the source database name
+##'   and the value being the instance for this database
+##'   (e.g. \code{source = "production"}).
+##'
```
weshinsley: See comments on similar text in db.R

R/db.R
```diff
@@ -27,6 +27,16 @@
...
+##'   select the instance, where multiple instances are configured.
+##'   The options here can vary - use a single \emph{unnamed}
+##'   character string to indicate an instance to match.  If given,
+##'   then this name must be present in all databases that support
+##'   instances, and will be ignored by all that do not.  Otherwise,
+##'   provide \emph{named} character vector with the name being the
+##'   source database name and the value being the instance for this
+##'   database (e.g. \code{source = "production"}).
+##'
```
weshinsley: Hmm... I am probably a bit rusty on how this works - but I'm finding this para a bit hard to get my head around. It may be easiest to refer to the vignette and have an expanded example, as we probably don't want masses of documentation right here...

I think it's the middle sentence about whether databases support instances that, and what happens if the instance is ignored, that is getting me...

For the VIMC scenario:
So different "databases" might be (eg, source (montagu) and annex?). In that case, annex wouldn't "support" instances (or rather, we wouldn't specify any - because we've only got one of it), but montagu could have instances production, science, uat and latest as instances, in the config file.

So: if we had source (montagu) and annex, we could specify instance="uat", which would do something for montagu, but be safely ignored for the one and only annex.

And if we did have separate instances on annex, then we could specify:
instance = c(source = "uat", annex = "staging_annex")

although then I think I've gone wrong somewhere with the sentence that says this is only used by type="source".

vignettes/orderly.Rmd
```diff
@@ -571,6 +571,38 @@ would look up the field `password` at the path
...
+        host: production.example.org
+        password: $PASSWORD_PRODUCTION
+      staging:
+        host: staging.example.org
+        password: $PASSWORD_STAGING
+    default_instance: $DEFAULT_INSTANCE
+```
+
+Here - staging and production have different hostnames (`production.example.org` and `staging.example.org`) and different passwords (retrieved using environment variables) and the default instance is set with another environnent variable (`$DEFAULT_INSTANCE`, which must be one of `production` or `staging`).  To switch between databases, you can set that variable, or pass the `instance` argument to `orderly::orderly_run` and friends, as:
+
```
weshinsley: Possibly, adding another database (eg annex) to this config file would be good - to demonstrate the multi database sort of configuration that gets mentioned earlier...

R/db.R
richfitz: Here is the vimc situation: https://github.com/vimc/montagu-reports/compare/vimc-3318 - once this PR is settled I can open a PR for this and get it going.

So we only have one annex instance and the named version of the instance argument would not be used.  It's probably overkill but I think gets us all the way to the most general solution.

Our database instances are going to be `production`, `science`, `uat` and `latest` corresponding to all the full copies of the db.  Annex always refers to whatever is configured on that machine (though we could do this with instances it's a bit awkward as we have one global annex on annex.montagu and a fake annex on uat which is used only by uat).

vignettes/orderly.Rmd
richfitz: I've expanded that a bit

weshinsley: Great - I think this (and the below) works really well for me.

PR #154, opened 2020-01-28
vimc-3375: Save random seed alongside metadata

richfitz:
As discussed with the coronavirus group, this would allow some (somewhat) automated workflows for reproducing stochastic analyses, even if `set.seed` is not used.

If we store `.Random.seed` then we can in theory restore the state of the RNG:

```r
> set.seed(1)
> rs <- .Random.seed
> runif(10)
 [1] 0.26550866 0.37212390 0.57285336 0.90820779 0.20168193 0.89838968
 [7] 0.94467527 0.66079779 0.62911404 0.06178627
> .Random.seed <- rs
> runif(10)
 [1] 0.26550866 0.37212390 0.57285336 0.90820779 0.20168193 0.89838968
 [7] 0.94467527 0.66079779 0.62911404 0.06178627
```

There's no support for automatically using this on rerun (and reports where this is important should use the `set.seed()` functionality, which is preferred) but we can look at restoring it when doing VIMC-3378

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: orderly
 Title: Lightweight Reproducible Reporting
-Version: 1.0.4
+Version: 1.0.5
 Description: Order, create and store reports from R.  By defining a
     lightweight interface around the inputs and outputs of an
     analysis, a lot of the repetitive work for reproducible research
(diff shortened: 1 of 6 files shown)


DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: orderly
 Title: Lightweight Reproducible Reporting
-Version: 1.0.4
+Version: 1.0.5
```
weshinsley: Just to confirm - 1.0.5 is correct if this is merged before #150

PR #156, opened 2020-02-05
vimc-3377: more flexible draft selection

richfitz:
This PR is intended to (eventually) replace the `draft: true` trick from orderly.yml's depends, by the ability to change where dependencies come from when running the report. This means that the underlying source does not need updating and we get fewer cases of reports being merged with draft: true sticking around

### DIFF
--- a/R/query.R
+++ b/R/query.R
@@ -190,25 +190,77 @@ orderly_find_name <- function(id, config, locate = FALSE, draft = TRUE,
   }
 }
 
+## This is only used in one place, and so we can be quite flexible in
+## how it is used.  It's a bit of a horror show, tbh, given how simple
+## what we want to achieve is.  This function is not an orderly API
+## function, and is used only in dependency resolution, so we can
+## update this later if needed.
+##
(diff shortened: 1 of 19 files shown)


R/query.R
```diff
@@ -190,25 +190,74 @@ orderly_find_name <- function(id, config, locate = FALSE, draft = TRUE,
...
-  if (!is.na(id) && file.exists(path_report)) {
-    return(path_report)
+
+  base_archive <- file.path(path_archive(config$root), name)
+  base_draft <- file.path(path_draft(config$root), name)
+
+  if (id == "latest") {
+    path <- NULL
+    if (search_archive) {
+      found <- orderly_latest(name, config, draft = FALSE, must_work = FALSE)
```
JamesThompson1729: Why don't we pass *must_work* to *orderly_latest* here (and below)?

richfitz: because either branch could fail (there could conceivably be no drafts but some archive reports) so the must_work is dealt with at the end of the function after trying (perhaps) both possibilities

PR #158, opened 2020-02-10
VIMC-3411: Validate parameters early

richfitz:
This PR fixes a problem that @sangeetabhatia03 and @imai-n found - if the user provides invalid parameters (from the point of view of orderly's data model) then the error is only thrown when the report is commited, and then with an incomprehensible message.  Here, we validate the parameters before running, and provide better messages that can be actioned

### DIFF
--- a/tests/testthat/test-db.R
+++ b/tests/testthat/test-db.R
@@ -103,10 +103,6 @@ test_that("avoid unserialisable parameters", {
   skip_on_cran_windows()
   path <- prepare_orderly_example("parameters", testing = TRUE)
   t <- Sys.Date()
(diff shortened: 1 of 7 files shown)


tests/testthat/test-db.R
```diff
@@ -103,10 +103,6 @@ test_that("avoid unserialisable parameters", {
   skip_on_cran_windows()
   path <- prepare_orderly_example("parameters", testing = TRUE)
   t <- Sys.Date()
-  id <- orderly_run("example", parameters = list(a = t, b = TRUE, c = "one"),
```
richfitz: This is now validated earlier, and the remaining tests never relied on a report being run

PR #161, opened 2020-02-18
vimc-3386: more informative resolution of environment variables

richfitz:
This solves a problem where (say) a remote is configured like:

```
remote:
  main:
    driver: orderlyweb::orderlyweb_remote
    args:
      host: orderly.example.com
      port: 443
      token: $VAULT_AUTH_GITHUB_TOKEN
```

and the environment variable is not set.  At present the user just gets an error saying

```
> orderly_pull_dependencies("explore_delays", remote = "testing")
Error: 'token' must be a scalar
> traceback()
15: stop(sprintf("'%s' must be a scalar", name), call. = FALSE)
14: assert_scalar(x, name)
13: assert_scalar_character(token)
12: orderlyweb_token_constant(token)
11: .subset2(public_bind_env, "initialize")(...)
10: R6_orderlyweb_api_client$new(host, port, token, name = name, 
        https = https, prefix = prefix, api_version = api_version, 
        insecure = insecure, verbose = verbose)
9: orderlyweb_api_client(...)
8: orderlyweb(host = host, port = port, token = token, https = https, 
       prefix = prefix, name = name)
7: .subset2(public_bind_env, "initialize")(...)
6: R6_orderlyweb_remote$new(host, port, token, https, prefix, name)
5: (function (host, port, token, https = TRUE, prefix = NULL, name = NULL) 
   {
       R6_orderlyweb_remote$new(host, port, token, https, prefix, 
           name)
   })(host = "ncov.dide.ic.ac.uk", port = 1443L, token = NULL, name = "testing")
4: do.call(driver, args)
3: load_remote(remote, config)
2: get_remote(remote, config)
1: orderly_pull_dependencies("explore_delays", remote = "testing")
```

which is not helpful and not actionable.  Now the error would look like:

```
> orderly_pull_dependencies("explore_delays", remote = "testing")
Error: Environment variable 'VAULT_AUTH_GITHUB_TOKEN' is not set
	(used in orderly_config.yml:remote:testing:args:token)
```

which points much more effectively to the error!  This is added to all environment variables that are resolved from the configuration.

### DIFF
--- a/tests/testthat/test-util.R
+++ b/tests/testthat/test-util.R
@@ -46,15 +46,12 @@ test_that("resolve_env", {
   set.seed(1)
   v <- paste(sample(c(LETTERS, 0:9, "_"), 20, replace = TRUE), collapse = "")
   vv <- paste0("$", v)
-  expect_identical(resolve_env(c(x = v)), list(x = v))
+  expect_identical(resolve_env(c(x = v), "loc"), list(x = v))
   expect_error(
-    resolve_env(c(x = vv)),
-    sprintf("Environment variable '%s' is not set.*used in x", v))
-  expect_error(
-    resolve_env(c(x = vv), name = "foo"),
-    sprintf("Environment variable '%s' is not set.*used in foo:x", v))
+    resolve_env(c(x = vv), "loc"),
+    sprintf("Environment variable '%s' is not set.*used in loc:x", v))
   expect_identical(
-    withr::with_envvar(setNames("value", v), resolve_env(c(x = vv))),
+    withr::with_envvar(setNames("value", v), resolve_env(c(x = vv), "loc")),
     list(x = "value"))
 })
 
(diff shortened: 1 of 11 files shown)


tests/testthat/test-util.R
```diff
@@ -46,24 +46,27 @@ test_that("resolve_env", {
...
+    list(x = "value"))
 })
 
 
 test_that("resolve_env skips non-scalars", {
   set.seed(1)
   v <- paste(sample(c(LETTERS, 0:9, "_"), 20, replace = TRUE), collapse = "")
   vv <- paste0("$", v)
-  expect_identical(resolve_env(v), list(v))
+  expect_identical(resolve_env(c(x = v)), list(x = v))
```
weshinsley: Note mostly to self - so, `used_in` (no default value) is not specified when calling resolve_env here - but it should not be needed in the test. Hence the test passes, and if `used_in` were required at some point during resolve_env, the test would correctly fail.

weshinsley: I think this is the only time you call resolve_env in such a way though, so looked like it needed exploring at first glance.

richfitz: thanks, I had thought I'd caught all of these

PR #167, opened 2020-03-02
VIMC-2929: Remove custom fields from main orderly table

richfitz:
Comment 2020: This removes the last bit of the pre-proper-schema orderly db - adding custom fields to the report_version table.  The schema version number increases (which triggers a db rebuild) but this should have minimal visibility to users.

This should be tested on uat with montagu pointing at this version to make sure nothing breaks.

Comment 2022: This has to be merged/deployed in lock step with https://github.com/vimc/orderly-web/pull/476 (and waiting for OW master build to be fixed before that build can pass!) Then will need to rebuild `orderly.server`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: orderly
 Title: Lightweight Reproducible Reporting
-Version: 1.4.9
+Version: 1.4.10
 Description: Order, create and store reports from R.  By defining a
     lightweight interface around the inputs and outputs of an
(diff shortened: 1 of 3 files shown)


hillalex:
Reviving this. This and https://github.com/vimc/orderly-web/pull/476 have to be tested and deployed together

richfitz:
The date on that PR offers some suggestions why it dropped out of view

hillalex (approved):

Issue #169, opened 2020-03-02
document that orderly_develop_start() also loads sources and packages


jeffeaton:
I was hoping that `orderly_deveop_start()` would also load my packages and sources so that I am all ready to go in my session once I run it, which I think it does, which is great.

I couldn't find this clearly stated in the documentation `?orderly_develop_start` or in the `orderly` getting started vignette, and so had to do some testing to confirm it was loading those.

Suggestion to add this clearly in the documentation that it will do this bit for you.

Thanks,
Jeff

jeffeaton:
Another comment on `orderly_develop_start()`: it would be nice to be able to execute this without specifying an artefact yet as I might only think of the artefacts I need as I develop the script.

```
> orderly::orderly_develop_start()
[ name       ]  fit-model
Error in recipe_read_check_artefacts(info$artefacts, filename, path) : 
  At least one artefact required
```

Not sure the consequences of allowing this under the hood. It's easy enough to put a 'dummy' file in for artefact if allowing no artefacts at this stage is problematic.

richfitz:
This was something I was wondering about too, yes, thanks

PR #174, opened 2020-03-03
VIMC-3503: Add usethis-style orderly_use_resource

### LINKED ISSUE #171: add function to add resource to a report
A function to automate the process of adding a resource to a report would be helpful, similar to `usethis::use_*` functionality. It would do the following:

- copying a file into the `src/<report>` directory (possibly with options to put it elsewhere as appropriate).
- add the file to `orderly.yml`.

Thanks,
Jeff

richfitz:
See #171 where @jeffeaton asks helper functions.  This is something also discussed in #160 (VIMC-3404).  Resources is probably the easiest case because it's an easy part of the yaml to edit (it's just an array) and because the files are already there. Unlike the requested spec, this PR does not copy the file in, though we could look to add that later perhaps.

Editing yaml while preserving non-syntactic properties is hard in general, so this PR implements some basic heuristics to hopefully do an ok job of it.  It will not work well on pathalogical yaml, but should do ok on the likely simple yaml that people actually write.

Fixes #171


richfitz:
Fixes #171

jeffeaton:
Thanks, this looks good.  I've just been using it to add a new report. Two questions so far:

* Why is `use_*(..., locate = FALSE)` default? It would be nice for `locate = TRUE` so that this would 'just work':

```
> library(orderly)
> orderly_use_package("naomi")
Error in orderly_config_get(root, locate) : Invalid input
```
* Then, similar to comment on #169, it would be nice for this to work before I have specified my artefact: 
```
> orderly_use_package("naomi", locate = TRUE)
Error in recipe_read_check_artefacts(info$artefacts, filename, path) : 
  At least one artefact required
```

and possibly before I have specified my script (after updating `orderly.yml` to have an artefact): 
```
> orderly_use_package("naomi", locate = TRUE, edit = FALSE)
Error: '/Users/jeff/Dropbox/Documents/Code/R/naomi-orderly-example/src/output-data-pack/orderly.yml:script' must be a scalar
```

I suspect that these are better to be addressed under a separate issue though.

* Then when I have an `orderly.yml` with a script and an artefact, I am getting an error: 
```
> orderly_use_package("naomi", locate = TRUE)
Error in tmp[!grepl("^\\s*(#.*)?$", tmp)][[1L]] : subscript out of bounds
```
The reason seems to be because I have `packages:` but nothing following it in my `orderly.yml`, which is probably invalid?:

```
script: datapack-export.R

artefacts:
  - data:
      description: PEPFAR Data Pack export file
      filenames: naomi_datapack.csv

packages:
```
This doesn't seem a super crazy thing for a user to have done -- if there's a way to either allow this, or give a more descriptive error message for why it has gone wrong.

-----
After I removed the `packages:` from `ordely.yml` to be;
```
script: datapack-export.R

artefacts:
  - data:
      description: PEPFAR Data Pack export file
      filenames: naomi_datapack.csv
```
it goes beautifully:
```
> orderly_use_package("naomi", locate = TRUE)
Changes to 'src/output-data-pack/orderly.yml'
6 |       filenames: naomi_datapack.csv
7 | 
8 | packages:
9 |   - naomi
Write changes to file?  

1: no
2: yes

Selection: 2
Writing to 'src/output-data-pack/orderly.yml'
```
The prompt is a very nice feature.

richfitz:
Thanks Jeff - very useful.  Comments below (quoting an abbreviated version of your comment for context)

> * Why is `use_*(..., locate = FALSE)` default? It would be nice for `locate = TRUE` so that this would 'just work':

Fixed now in this PR

> * Then, similar to comment on #169, it would be nice for this to work before I > have specified my artefact: 
> ```
> > orderly_use_package("naomi", locate = TRUE)
> Error in recipe_read_check_artefacts(info$artefacts, filename, path) : 
>   At least one artefact required
> ```
> 
> and possibly before I have specified my script (after updating `orderly.yml` to > have an artefact): 
> ```
> > orderly_use_package("naomi", locate = TRUE, edit = FALSE)
> Error: '/Users/jeff/Dropbox/Documents/Code/R/naomi-orderly-example/src/output-data-pack/orderly.yml:script' must be a scalar
> ```

Fixed in #176, VIMC-3510 (you reported this also in #169)

> * Then when I have an `orderly.yml` with a script and an artefact, I am getting > an error: 
> ```
> > orderly_use_package("naomi", locate = TRUE)
> Error in tmp[!grepl("^\\s*(#.*)?$", tmp)][[1L]] : subscript out of bounds
> ```
> The reason seems to be because I have `packages:` but nothing following it in my `orderly.yml`, which is probably invalid?:

> ```
> script: datapack-export.R
> 
> artefacts:
>   - data:
>       description: PEPFAR Data Pack export file
>       filenames: naomi_datapack.csv
> 
> packages:
> ```
>
> This doesn't seem a super crazy thing for a user to have done -- if there's a way to either allow this, or give a more descriptive error message for why it has gone wrong.

I can't replicate this - could you please open another ticket with full details to reproduce.  Adding a line `packages:` to the example that is produced by `orderly::orderly_example("minimal")` I get:

```
> writeLines(readLines("orderly.yml"))
data:
  dat:
    query: SELECT name, number FROM thing
script: script.R
artefacts:
  staticgraph:
    description: A graph of things
    filenames: mygraph.png
packages:
> orderly::orderly_develop_start()
[ name       ]  example
[ data       ]  source => dat: 20 x 2
```

jeffeaton:
Thanks very much. New issue for the last created in #177.

tests/testthat/test-util.R
```diff
@@ -698,3 +698,93 @@ test_that("random_seed", {
...
+
+
+test_that("insert into files", {
+  text <- c("a", "b", "c")
+  value <- c("x", "y")
+  path <- tempfile()
+  writeLines(text, path)
+
+  expect_message(
+    str <- capture.output(
```
r-ash: This is a new thing for me and shows up something I haven't seen before. When I've done things like this where I want to write tests about the messages printed and the result returned I've used `testthat::evaluate_promise` which I think is doing something similar to here but it returns the `result`, `message` and the `output`. So I wasn't aware the was a difference here.

What is the difference between `output` and `result`? Result is the return from the function and the `output` is what is coming from call to `cat`. Perhaps I don't quite get how `cat` works. How is `cat` different from a `message` or from `print`?

R/util.R
```diff
@@ -730,3 +730,96 @@ random_seed <- function(envir = globalenv()) {
...
+  ##
+  ## TODO: Better behaviour for the key: value pair where value could
+  ## have been a list.
+  re <- sprintf("^%s\\s*:", name)
+  start <- grep(re, text)
+  if (length(start) == 0L) {
+    return(list(name = name, exists = FALSE, block = FALSE))
+  }
+
+  re <- sprintf("^%s\\s*:\\s*(#.*)?", name)
```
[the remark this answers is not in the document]
r-ash: Oh I see that line lower down has been removed since I wrote this - feel free to ignore.

tests/testthat/test-util.R
richfitz: I was looking for `testthat::evaluate_promise` but could not find it - I'll rewrite to use that as it will be clearer.

cat writes to stdout, message to stderr, print *probably* writes to stdout, but it relies on underlying cat/message calls to do the printing

richfitz:
Thanks @r-ash - just made that change now. Thanks for the reminder of that function, I really couldn't remember what it was called...

PR #179, opened 2020-03-04
VIMC-3514: Basic support for tags

richfitz:
This PR adds very basic support for tags.  At the moment, the main thing it implements is the db structure (new tables `tag` and `report_version_tag`).  Nothing useful can be done with tags within orderly.  The demo data set has tags added to two reports for use in orderlyweb testing.


richfitz:
@r-ash - what do you think about passing that through orderly_run too?

PR #185, opened 2020-03-10
vimc-3538: Add query interface

richfitz:
This PR introduces a basic query interface for searching for reports that match criteria based on parameters and tags, which can be used directly `orderly::orderly_search` or when declaring dependencies (VIMC-3538)

I would like to make sure that we get this right!  The PR breaks into two basic chunks and could be refactored into these chunks, though that makes it a bit more abstract.

**In the first**, we define a query language (`query_search.R`) - this has essentially no impact on existing code, and we can change this in anyway that we want.

**The second part** is uglier and infects `develop.R`, `recipe_read.R`, and `recipe_read_depends.R` where we wire this up to be useful for declaring dependency.  Unfortunately this requires passing `parameters` throughout.

Note that this does not work for resolution against a remote.

~(edit: I see there are coverage holes that need filling before merge too, but they should not affect the logic)~

(edit to remove the following; after talking with Rob we decided to start off verbose and unambiguous)

> One wrinkle with the query language as it stands is that parameters are ambiguous, so:
> 
> ```
> nmin < nmin
> ```
> 
> always takes the first argument as the parameter used in the report and the second as the one passed in via `parameters`.  We could require
> 
> ```
> parameter:nmin < nmin
> ```
> 
> always, or where ambiguous.  As I write this, I am now doubting if report parameters will be substituted correctly if used on the rhs...


R/query_search.R
```diff
@@ -0,0 +1,414 @@
...
+parse_query_filter <- function(expr, parameters) {
+  if (!is.call(expr)) {
+    stop("Expected an expression")
+  }
+
+  if (is_call(expr, ":") && length(expr) == 3L) {
+    type <- expr[[2L]]
+    stopifnot(is.symbol(type))
+    if (!identical(type, quote(tag))) {
+      stop(sprintf("Invalid query expression '%s' requires operator",
```
r-ash: I think this will be clearer about what the problem is if we pass the `type` to the error message.

Something like
`Invalid query expression 'z:thing', operator 'z' not supported - supported operators: 'tag', 'parameter'`

r-ash: Ah I see this captures when you pass `paramter:thing` in too without an operator like `>`.

What I am trying to say is if we do something like
```
orderly::orderly_search("something:123", "other", root = root)
```
We get an error

```
 Error: Invalid query expression 'test:123' requires operator
```
which is a bit misleading. I can see people trying to do that with some other search term other than `tag` or `parameter` but perhaps that is unlikely to happen?

tests/testthat/test-query-search.R
```diff
@@ -0,0 +1,267 @@
...
+
+  res <- parse_query('latest(a > 1 && (b == "value" || b == "other"))', NULL)
+  expect_true(res$latest)
+  expect_equal(res$use, list(parameter = TRUE, tag = FALSE))
+  expect_equal(res$expr,
+               quote(parameter[["a"]] > 1 &&
+                     (parameter[["b"]] == "value" ||
+                      parameter[["b"]] == "other")))
+
+  res <- parse_query('a > 1 && tag:weekly')
```
r-ash: Any expectations for this case?

R/query_search.R
```diff
@@ -0,0 +1,414 @@
...
+## * parameter:id == 1
+## * tag:weekly
+parse_query_filter <- function(expr, parameters) {
+  if (!is.call(expr)) {
+    stop("Expected an expression")
+  }
+
+  if (is_call(expr, ":") && length(expr) == 3L) {
+    type <- expr[[2L]]
+    stopifnot(is.symbol(type))
```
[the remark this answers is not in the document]
richfitz: no reason, have standardised on `is.symbol` as it's most commonly used and most explicit

R/query_search.R
richfitz: Thanks - this was very useful. Have updated now so errors are like

```
> orderly::orderly_search("something:123", "other", root = path)
Error: Invalid namespaced query element 'something:123'; expected symbol for key
> orderly::orderly_search("something:abc", "other", root = path)
Error: Query namespace (used as 'something') must be one of 'tag', 'parameter'
```

tests/testthat/test-query-search.R
richfitz: Oops, added some

R/query_search.R
```diff
@@ -0,0 +1,467 @@
...
+##' The syntax for tags is simpler, one uses \code{tag:tagname} to
+##' test for presence of a tag called "tagname".
+##'
+##' Search queries can be joined by \code{&&} and \code{||} and
+##' grouped using parentheses, these groups (or tags) can be negated
+##' with \code{!}, so a complicated query expression might look like:
+##'
+##' \preformatted{
+##' (parameter:fruit == "apple" && !tag:weekly) || paramerter:fruit == "banana"
+##' }
```
weshinsley: Typo on the second parameRTer

R/query_search.R
```diff
@@ -0,0 +1,467 @@
...
+##' # or exclude them
+##' orderly::orderly_search("!tag:plot", "other", root = root)
+##'
+##' # or combine that with the presence/absence of a tag
+##' orderly::orderly_search("parameter:nmin > 0.15 && !tag:plot",
+##'                         "other", root = root)
+##'
+##' # Use latest() over a query to find the latest report matching the
+##' # query expression.
+##' orderly::orderly_search("latest(parameter:nmin > 0.15)",
```
weshinsley: You haven't got any examples of `[!]is.null(parameter:thing)` - but possibly not needed!

R/query_search.R
```diff
@@ -0,0 +1,467 @@
...
+  }
+
+  namespace <- "parameter"
+  rel <- as.character(expr[[1L]])
+  if (rel == "is.null") {
+    stopifnot(length(expr) == 2L)
+    res <- parse_query_namespace(expr[[2L]])
+    if (!identical(res$namespace, "parameter")) {
+      stop(sprintf(
+        "In '%s', query namespace must be 'parameteter' but found '%s'",
```
weshinsley: Typo... parameter is hard to type!

R/query_search.R
```diff
@@ -0,0 +1,467 @@
...
+        paste("Query parameter '%s' not found in supplied parameters",
+              "did you mean 'parameter:%s'?"),
+        name, name), call. = FALSE)
+    }
+    ret <- list(namespace = NULL, key = value)
+  } else if (is_call(x, ":")) {
+    ret <- parse_query_namespace(x)
+    if (ret$namespace != "parameter") {
+      stop(sprintf(
+        "In '%s', query namespace must be 'parameteter' but found '%s'",
```
weshinsley: Parameter typo again!

tests/testthat/test-query-search.R
```diff
@@ -0,0 +1,388 @@
...
+    "Query namespace (used as 'something') must be one of 'tag', 'parameter'",
+    fixed = TRUE)
+  expect_error(
+    parse_query("something:abc > 1", NULL),
+    "Query namespace (used as 'something') must be one of 'tag', 'parameter'",
+    fixed = TRUE)
+  ## TODO:
+  expect_error(
+    parse_query("tag:abc > 1", NULL),
+    "In '.+', query namespace must be 'parameteter' but found 'tag'")
```
weshinsley: Parameter typo - to match query_search:354 or 395

tests/testthat/test-query-search.R
```diff
@@ -0,0 +1,388 @@
...
+})
+
+
+test_that("is.null requires a namespace", {
+  expect_error(
+    parse_query("is.null(x)"),
+    "Expected namespaced query element but recieved 'x'")
+  expect_error(
+    parse_query("is.null(tag:thing)"),
+    "In 'is.null(tag:thing)', query namespace must be 'parameteter' but found 'tag'",
```
weshinsley: Parameter typo - to match query_search:354 or 395

R/query_search.R
```diff
@@ -0,0 +1,467 @@
...
+      call. = FALSE)
+  }
+  ret
+}
+
+
+parse_query_namespace <- function(expr) {
+  if (!is_call(expr, ":")) {
+    stop(sprintf(
+      "Expected namespaced query element but recieved '%s'",
```
weshinsley: recEIved

tests/testthat/test-query-search.R
```diff
@@ -0,0 +1,388 @@
...
+    list(latest = FALSE,
+         use = list(parameter = TRUE, tag = FALSE),
+         expr = quote(1 > parameter[["x"]])))
+})
+
+
+test_that("is.null requires a namespace", {
+  expect_error(
+    parse_query("is.null(x)"),
+    "Expected namespaced query element but recieved 'x'")
```
weshinsley: recIEved - to match query_search.R:411

R/query_search.R
```diff
@@ -0,0 +1,467 @@
...
+##' \preformatted{
+##' (parameter:fruit == "apple" && !tag:weekly) || paramerter:fruit == "banana"
+##' }
+##'
+##' For clarity, parameters may be prefixed with \code{parameter:}
+##' (so, \code{parameter:fruit} in the above), though this is
+##' optional.  In the documentation and error messages we may refer to
+##' the left-hand-side of \code{:} as a "namespace".  At this point
+##' the only supported namespaces are \code{tag} and \code{parameter}.
+##'
```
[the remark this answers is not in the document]
richfitz: sorry, out of date - this is now compulsory, as it did in fact cause ambiguity. I've updated docs now

R/query_search.R
```diff
@@ -0,0 +1,467 @@
...
+##' # We can look for tags
+##' orderly::orderly_search("tag:plot", "other", root = root)
+##'
+##' # or exclude them
+##' orderly::orderly_search("!tag:plot", "other", root = root)
+##'
+##' # or combine that with the presence/absence of a tag
+##' orderly::orderly_search("parameter:nmin > 0.15 && !tag:plot",
+##'                         "other", root = root)
+##'
```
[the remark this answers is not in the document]
richfitz: I've added a warning telling people not to do that :)

R/query_search.R
```diff
@@ -0,0 +1,477 @@
...
+##' grouped using parentheses, these groups (or tags) can be negated
+##' with \code{!}, so a complicated query expression might look like:
+##'
+##' \preformatted{
+##' (parameter:fruit == "apple" && !tag:weekly) || parameter:fruit == "banana"
+##' }
+##'
+##' Be careful of comparing floating point numbers with \code{==} or
+##' \code{!=} may not always return what you expect (for example
+##' \code{sqrt(3)^2 == 3} is \code{FALSE}).
```
weshinsley: [because they] may not always return

PR #189, opened 2020-03-16
vimc-3569: de-lint

richfitz:
No new features, this just picks off a bunch of small linty bits and sets up codefactor, as we did with dettl


R/file_store.R
```diff
@@ -110,18 +108,18 @@ R6_file_store <- R6::R6Class(
 
 ## This is directly from storr and needs to have a PascalCase class
```
r-ash: Out of date comment now?

PR #190, opened 2020-03-16
VIMC-3570: New "workflows" vignette

richfitz:
This PR is an attempt at getting some things that exist only in my head out and documented.  I've already highlighted some sections that require some additional work with diagrams etc but I am sure that there is more.  All comments would be welcome

* add something about how drafts and local archives fit into this


vignettes/workflows.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+Backup is the responsibility of the researcher; if the computer is backed up, or if the archive lives on a service like Dropbox or OneDrive, this may be sufficient.
+
+### Why not commit the archive to git?
+
+If your reports are very small, then this may work for you.  However, git becomes slow to use when many large files have been added to it, and most hosting platforms, including GitHub will prevent addition of files over a certain size and may complain about overall repositories reaching a certain size.  Cloning the source tree will become increasingly painful, and you may become wary of even *running* a report because of the impact it may have on the above problems.  In our opinion, you should store your archive outside of git and use an alternative method to back it up and distribute it.
+
```
weshinsley: Reading this para made me think about the philosophy of reproducible reports...

The source code to generate a single graph might well be far smaller in size than the image file that the code produces. For the source code to create a hundred similar graphs might involve the addition of a couple of lines of loop code, whereas the combined size of the image files the code produces might be a hundred times larger than before. Orderly focuses on storing the code that reproduces a report, rather than storing the report itself, which is a good idea both in terms of keeping storage needs manageable, and also in encouraging the principle of reproducibility in reporting analyses.

vignettes/workflows.Rmd
```diff
@@ -0,0 +1,210 @@
...
+## Small group, sharing via shared disk
+
+As soon as more than one person is involved, things get more tricky.  We need to ensure that everybody has a few of the same set of archive reports, and over the same source tree, yet can work on their own work somewhat independently.
```
weshinsley: few -> view ?

vignettes/workflows.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+The researchers use a common "source tree" which is under version control with git, and kept on GitHub.  Nobody may commit to master and all changes to master are introduced via a ["Pull Request"](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests) (other hosted source control platforms have equivalents, for example [GitLab merge requests](https://docs.gitlab.com/ee/user/project/merge_requests/)).  Everybody has this cloned locally, and the git repository does not include `archive` or `source`.  With this setup it is easy to ensure that different members of the team are looking at the same source code, and `orderly` will track that as part of its metadata (git's branch, sha and status).
+
```
weshinsley: Do you mean: the git repository does not include `archive` or `draft` ?   (but it does include `src`)

vignettes/workflows.Rmd
```diff
@@ -0,0 +1,210 @@
...
+However, we still must synchronise the reports that have been run.  Of course everyone can run their own copy of a report and build off of that but this poses several potential problems:
+
+* if the report is stochastic, the results will likely differ
```
weshinsley: Would "non-deterministic" be better than "stochastic" ? Or "a stochastic instance" ...

vignettes/workflows.Rmd
```diff
@@ -0,0 +1,210 @@
...
+* if the report depends on an external data source (a SQL database, web resource, etc) the results may differ
+* if underlying package versions have changed, the results may differ
+* if the report is long-running, this is time-consuming
+* everyone will end up with a different report id, so it is difficult to know that we're looking at the same "effective" report when results are shared
+
+What we want is that after a particular version of a report to have been run, it is easy for other people to start depending on it.  To achieve this we need a method for getting the archive from one person's computer to another.
```
weshinsley: "to have been run" -> "has been run" maybe? Slightly too many words...

vignettes/workflows.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+For this discussion we will assume two researchers ["Alice" and "Bob"](https://en.wikipedia.org/wiki/Alice_and_Bob) rather than "Researcher A" and "Researcher B".
+
+Alice develops a report `report-a` on her computer by running `orderly::orderly_new("report-a")` and working with `orderly::orderly_develop_start()` and `orderly::orderly_run()` until she is happy with the results.  She then creates a pull request, and Bob approves it (probably after checking that the report runs on his machine too) and merged it.
```
weshinsley: at the end, merged -> merges

vignettes/workflows.Rmd
```diff
@@ -0,0 +1,210 @@
...
+```yaml
+ORDERLY_PATH_DEFAULT_REMOTE: /Users/alice/Dropbox/ourorderly
+```
+
+and for Bob it might look like
+
+```yaml
+ORDERLY_PATH_DEFAULT_REMOTE: c:/Bob/Documents/Dropbox/ourorderly
+```
+
```
weshinsley: I am wondering here whether the word "orderly" is confusing - on lines 69, "shared copy of orderly", and the two examples of alice and Bob's Dropbox being called ourorderly - is it better to call that something to do with reports/archives - eg, the remote being referred to as shared_archive, and the local being my_archive or local_archive

vignettes/workflows.Rmd
```diff
@@ -0,0 +1,210 @@
...
+With this done, Bob can get a copy of `report-a` by running
+
+```r
+orderly::orderly_pull_archive("report-a")
+```
+
+which will copy the latest version of the report from the shared archive into their personal archive.  A this point, Bob has *exactly* the copy of `report-a` that was run on the shared archive.  In the same way that Alice and Bob can exchange git sha hashes to ensure that they are on the same version of the source tree, they can now exchange report ids and ensure that they are looking at the same output.
```
weshinsley: What's your favourite way of showing the hash? `git rev-parse HEAD`?

vignettes/workflows.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+If you are working on a report that has a dependency (say `report-b`, which depends on `report-b`), you can automatically pull its dependencies by running
```
weshinsley: Presumably the first should be `report-a` ?

vignettes/workflows.Rmd
```diff
@@ -0,0 +1,210 @@
...
+      token: $GITHUB_TOKEN
+```
+
+Once this is done, both Alice and Bob can run reports *on* that server, for example
+
+```r
+orderly::orderly_run_remote("report-a", remote = "default")
+```
+
+**ADD A GIF**
```
weshinsley: Todo!

vignettes/workflows.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+this will print to screen the log, just as if it was running locally, though it is in fact running on a fresh R session remotely.  After being run, this report is available to everyone and can be pulled as above
```
weshinsley: as if it *were* running locally...

vignettes/workflows.Rmd
```diff
@@ -0,0 +1,210 @@
...
+      port: 11443
+      token: $GITHUB_TOKEN
+  production:
+    driver: orderlyweb::orderlyweb_remote
+    args:
+      host: orderly.example.com
+      token: $GITHUB_TOKEN
+```
+
+Now, as part of the pull request review procedure, the reviewer will run the report on the staging environment, which ensures that it will likely work without error on the production environment.
```
weshinsley: Something about the words "ensures" and "likely" being close together...
perhaps "should ensure that it will work without error" ?

vignettes/workflows.Rmd
```diff
@@ -0,0 +1,210 @@
...
+      ...
+```
+
+## Dependency resolution
+
+As soon as there are multiple archives, resolving dependencies becomes more challenging.  If Alice has two reports that depend on each other, she might want to use the most recent versions on her machine, or the most recent versions *as seen by one of her remotes* as that is where the report will eventually be run.
+
+To allow this, when running `orderly::orderly_run`, Alice can provide the argument `remote`, indicating which of her remotes she would like to use for dependency resolution.
+
+**GIVE A CONCRETE EXAMPLE HERE, WITH DIAGRAM**
```
weshinsley: Yep - I think this is where it's easy to feel the slight plunge of "euuu, this is getting messy to think about"

vignettes/workflows.Rmd
richfitz: I've gone now with `our-project` as it avoids pretty much any overloading

vignettes/workflows.Rmd
richfitz: yes, it's madness that this is not easier to do

vignettes/workflows.Rmd
richfitz: other way around in fact

vignettes/workflows.Rmd
richfitz: Ah, the rare form of the English subjunctive

richfitz:
(Further work pushed to https://mrc-ide.myjetbrains.com/youtrack/issue/VIMC-3759)

PR #195, opened 2020-03-30
VIMC-3605: Add orderly_batch function for batch running reports

r-ash:
This PR will
* Add a function `orderly_batch` for running a report with multiple sets of parameters

This just creates a `batch_id` and writes that into the metadata of the output rds file. Coming next
* CLI support
* Writing to DB on commit


R/batch.R
```diff
@@ -0,0 +1,33 @@
...
+##' 
+##' @examples
+##' 
+##' path <- orderly::orderly_example("demo")
+##' params <- data.frame(nmin = c(0.2, 0.25))
+##' ids <- orderly::orderly_run("other", params, root = path)
+orderly_batch <- function(name = NULL, parameters = NULL, ...) {
+  ## Assert each column named? or a test at least if not to see what happens
+  batch_id <- ids::random_id()
+  ids <- lapply(seq_len(nrow(parameters)), function(row) {
```
r-ash: I am sure you have shown me a better way to loop over rows in a type safe way. I couldn't remember what it was or where - can you remember?

richfitz: there is no great way of doing this - I'd be inclined to make a utility function df2list:

```
df2list <- function(df) {
  unname(split(df, seq_len(nrow(df)))
}
```

and then vcapply over that, or you could vcapply over what you have already

richfitz:
(also the example is broken :) )

R/batch.R
r-ash: Nice - that looks tidy. Thanks

PR #197, opened 2020-04-01
VIMC-506: Refactor recipe class

richfitz:
This PR was not a great deal of fun to write, but it tidies away some technical debt.  There is at least one more of these to go (following on from #193) - VIMC-3611.

This adds no functionality, but centralises and standardises the orderly.yml validation with the general hope of eventually supporting plugins.  The logic remains the same though there are few bits patched up in the test.

We end up with a class that is a lot like the config one - it's mostly for collecting up data.  Not sure that this is a great design, but we can iterate over it more easily that lines of responsibility are more consistent now.

The actual running of a report is still a mess, unfortunately - that's VIMC-3611.  I've just patched this class up to support how it is currently used and will do that in the next PR on this line of work


R/config.R
```diff
@@ -330,7 +331,7 @@ config_validate_database <- function(database, filename) {
     return(NULL)
   }
 
-  assert_named(database, unique = TRUE, sprintf("%s:database", filename))
+  assert_named(database, unique = TRUE, sprintf("%s:rawbase", filename))
```
richfitz, line 334: this is a typo

PR #198, opened 2020-04-02
vimc-3608: Add support running batch from CLI

r-ash:
This PR will
* Expose new `orderly batch` command to CLI interface
* Called with comma separated list of params `param1=value1,value2 param2=value1,value2`

If this looks good I will look at adding a `--csv=PATH` arg in a separate ticket as we've discussed

As lots of the prep of params and the running here is similar to `orderly run` I have put those things together but ends up being some conditions. I think there is an argument for not touching the `orderly run` stuff and having some dupe code - what do you think is better here?


R/main.R
```diff
@@ -86,8 +87,27 @@ cli_args_process_run_parameters <- function(parameters) {
         "Invalid parameters %s - all must be in form key=value",
         paste(squote(parameters[err]), collapse = ", ")))
     }
-    value <- lapply(p, function(x) parse_parameter(x[[2L]]))
-    set_names(value, vcapply(p, "[[", 1L))
+    if (command == "batch") {
```
richfitz, line 90: I think this would be cleaner as two separate functions - one for batch and one for run

R/main.R
```diff
@@ -114,68 +134,96 @@ Options:
...
+    parameters <- x$options$parameters
+    id_file <- x$options$id_file
+    parameters <- x$options$parameters
+    print_log <- x$options$print_log
+    ref <- x$options$ref
+    fetch <- x$options$fetch
+    pull <- x$options$pull
+    message <- x$options$message
+
+    main_batch <- function() {
```
richfitz, line 160: This seems more complicated than ideal - I see you're trying to avoid repetition but  I think that comes at a cost of readability.  I think it's worth considering if we should be exposing all the same arguments to batch anyway - I think that id_file, commit should definitely disappear.  Some of the run complication exists to support the runner and we might want to look at that more holistically later.  Keeping the new part clean will help unpick that.

R/main.R
```diff
@@ -114,68 +134,97 @@ Options:
...
+    if (print_log) {
+      sink(stderr(), type = "output")
+      on.exit(sink(NULL, type = "output"))
+      ids <- main_batch()
+    } else {
+      log <- tempfile()
+      ## we should run this with try() so that we can capture logs there
+      ids <- capture_log(main_batch(), log)
+      dest <- (if (commit) path_archive else path_draft)(config$root)
+      ## TODO: Here all orderly logs get the history for all reports in the
```
r-ash, line 187: What do you think about this comment? Do we want the logs for all reports in the batch to be saved for every report within the batch?

orderly.Rproj
```diff
@@ -12,6 +12,8 @@ Encoding: UTF-8
 RnwWeave: Sweave
 LaTeX: pdfLaTeX
 
+StripTrailingWhitespace: Yes
```
richfitz, line 15: nice :)

R/main.R
richfitz: yeah, logs need to go into their individual directories

PR #200, opened 2020-04-02
VIMC-3617: Add support for redirecting orderly logs to a file

r-ash:
This PR will
* Add support for `orderly_run` and `orderly_commit` to write logs to a `orderly.log` file based on a config flag
* Use config flag in CLI run

### DIFF
--- a/R/main.R
+++ b/R/main.R
@@ -136,7 +136,7 @@ main_do_run <- function(x) {
   fetch <- x$options$fetch
   pull <- x$options$pull
   message <- x$options$message
-  
+
   if (print_log) {
     sink(stderr(), type = "output")
     on.exit(sink(NULL, type = "output"))
--- a/R/main.R
+++ b/R/main.R
@@ -136,38 +136,27 @@ main_do_run <- function(x) {
   fetch <- x$options$fetch
   pull <- x$options$pull
   message <- x$options$message
-
-  main_run <- function() {
-    if (pull) {
-      if (is.null(ref)) {
-        git_pull(config$root)
-      } else {
-        orderly_cli_error(
-          "Can't use --pull with --ref; perhaps you meant --fetch ?")
-      }
-    }
-    id <- orderly_run(name, parameters, root = config, id_file = id_file,
-                      instance = instance,
-                      ref = ref, fetch = fetch, message = message)
-    if (commit) {
-      orderly_commit(id, name, config)
-    }
-    id
-  }
-
+  
   if (print_log) {
     sink(stderr(), type = "output")
     on.exit(sink(NULL, type = "output"))
-    id <- main_run()
   } else {
-    log <- tempfile()
-    ## we should run this with try() so that we can capture logs there
-    id <- capture_log(main_run(), log)
-    dest <- (if (commit) path_archive else path_draft)(config$root)
-    file_copy(log, file.path(dest, name, id, "orderly.log"))
+    config$add_run_option("capture_log", FALSE)
   }
 
+  if (pull) {
+    if (is.null(ref)) {
+      git_pull(config$root)
+    } else {
+      orderly_cli_error(
+        "Can't use --pull with --ref; perhaps you meant --fetch ?")
+    }
+  }
+  id <- orderly_run(name, parameters, root = config, id_file = id_file,
+                    instance = instance,
+                    ref = ref, fetch = fetch, message = message)
   if (commit) {
+    orderly_commit(id, name, config)
     path_rds <- path_orderly_run_rds(
       file.path(config$root, "archive", name, id))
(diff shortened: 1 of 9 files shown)


R/recipe_commit.R
```diff
@@ -37,7 +37,11 @@ orderly_commit <- function(id, name = NULL, root = NULL, locate = TRUE) {
     }
   }
   workdir <- file.path(path_draft(config$root), name, id)
-  recipe_commit(workdir, config$root)
+  
+  capture <- inherits(root, "orderly_config") &&
```
richfitz, line 41: why use `root` here and not `config`?  should just be able to write:

```
capture_log <- isTRUE(config$get_run_option("capture_log"))
```

I think.

R/recipe_commit.R
```diff
@@ -37,7 +37,11 @@ orderly_commit <- function(id, name = NULL, root = NULL, locate = TRUE) {
     }
   }
   workdir <- file.path(path_draft(config$root), name, id)
-  recipe_commit(workdir, config$root)
+  
+  capture <- inherits(root, "orderly_config") && 
+    isTRUE(root$get_run_option("capture_log"))
+  logfile <- file.path(path_draft(config$root), name, id, "orderly.log")
+  conditional_capture_log(capture, logfile, recipe_commit(workdir, config$root))
```
richfitz, line 44: I think this will overwrite the actual run log?

R/recipe_run.R
```diff
@@ -143,17 +143,30 @@ orderly_run <- function(name = NULL, parameters = NULL, envir = NULL,
...
+    withr::with_envvar(
+      orderly_envir_read(config$root),
+      recipe_run(info, parameters, envir, config, echo = echo,
+                 instance = instance))
+  }
+  
+  capture <- inherits(root, "orderly_config") && 
+    isTRUE(root$get_run_option("capture_log"))
+  logfile <- tempfile()
+  info <- conditional_capture_log(capture, logfile, run(capture))
```
richfitz, line 169: I had wondered about writing it in a withr style:

```
info <- conditional_capture_log(capture, logfile, {
  info <- recipe_prepare(config, name, id_file, ref, fetch, message,
                         use_draft, parameters, remote, tags = tags,
                         batch_id = batch_id)

  if (capture) {
    dest <- path_draft(config$root)
    on.exit(file_copy(logfile, file.path(dest, name, info$id, "orderly.log")), 
            add = TRUE)
  }

  recipe_current_run_set(info)
  on.exit(recipe_current_run_clear(), add = TRUE)

  withr::with_envvar(
    orderly_envir_read(config$root),
    recipe_run(info, parameters, envir, config, echo = echo,
               instance = instance))
})
```

what do you think?

R/util.R
```diff
@@ -196,7 +196,12 @@ pasteq <- function(x, sep = ", ") {
 
 capture_log <- function(expr, filename) {
   ## nolint start
-  con <- file(filename, "w")
+  if (file.exists(filename)) {
+    mode = "a"
```
richfitz, line 200: why not `<-`? and why is nolint required?

richfitz: you can also just write 

```
mode <- if (file.exists(filename)) "a" else "w"
```

R/main.R
```diff
@@ -136,38 +136,27 @@ main_do_run <- function(x) {
...
   } else {
-    log <- tempfile()
-    ## we should run this with try() so that we can capture logs there
-    id <- capture_log(main_run(), log)
-    dest <- (if (commit) path_archive else path_draft)(config$root)
-    file_copy(log, file.path(dest, name, id, "orderly.log"))
+    config$add_run_option("capture_log", FALSE)
   }
 
+  if (pull) {
```
richfitz, line 147: we'll have to sort this mess out at some point too - seems bizarre that we pull here but fetch within the run command? What was I thinking?

R/recipe_run.R
r-ash: Yes I like that more - nicer to read than jumping up to the closure

R/recipe_commit.R
r-ash: Should be okay now we have the append mode when opening the file connection. This is tested too https://github.com/vimc/orderly/blob/16a58d1d01190b4d85edd727f022bde97257aac5/tests/testthat/test-run.R#L1010

R/util.R
r-ash: > why not `<-`? and why is nolint required?

Just being absent minded. I've got rid of the nolint stuff too.

R/recipe_commit.R
richfitz: sorry, I'd meant to delete that comment once I realised we're appending!

PR #205, opened 2020-04-06
VIMC-3636: Proper orderly buildkite support

richfitz:
This PR configures orderly correctly on buildkite and removes much of the TeamCity support.

It turns out that orderly was probably not the best starting point for moving things across to buildkite but with the 3 big images this is still one of the high-value targets.

This PR tidies up some loose ends:

* Emojis are added to the build steps which makes the output much nicer
* The orderly.server build is triggered by this build (only on master - see comment in pipeline.yml, as well as the [orderly.server pipeline.yml](https://github.com/vimc/orderly.server/blob/master/buildkite/pipeline.yml))
* We no longer push to the docker.montagu registry very much, nor reference it - I think we can entirely remove it at some point
* The build has been made safe to agent switching
* The TC build has been simplified down to just build the demo data as an artefact
* We also build the artefact here so that eventually the TC build (or the OrderlyWeb build if moved to buildkite) can use it
* The scripts run in verbose mode (+x) so that the logs are easier to deal with


buildkite/pipeline.yml
```diff
@@ -1,18 +1,35 @@
 steps:
-  - label: "Build"
+  - label: ":whale: Build"
```
r-ash, line 2: nice :) :whale:

PR #206, opened 2020-04-06
VIMC-3632: Add batch run info to orderly db

r-ash:
This PR will
* Add tables `report_version_batch` and `report_batch` to schema which get written to on commit

This seems quite small - I wonder if I have missed something? Mostly basing this off #179


NEWS.md
```diff
@@ -1,3 +1,7 @@
+# orderly 1.1.19
```
richfitz, line 1: should be 1.1.20 now. Consider rewording to reflect less the development process and more the user focus:

> Add support for running batches of orderly reports that differ in their parameters, available via `orderly::orderly_batch` and with the cli as `orderly batch` (VIMC-3603)

r-ash: Thanks, I've tidied that up

PR #209, opened 2020-04-15
vimc-3655: Helper function for getting orderly remote

richfitz:
This PR adds a minor feature that would have been helpful in writing some docs and will generally be helpful when tooling against orderly - we might want to interrogate the remote (using the orderlyweb client) but we don't want to have to wrap every endpoint (that creates effectively a circular dependency) or have to make the user create the client manually when we already know how to do this.

So here we just expose the 2 lines required to return the orderly remote object


R/remote.R
```diff
@@ -275,6 +275,34 @@ orderly_default_remote_get <- function(root = NULL, locate = TRUE) {
...
+##' @return The orderly remote, as described in
+##'   \code{orderly_config.yml} - if no remotes are configured, or if
+##'   the requested remote does not exist, an error will be thrown.
+##'
+##' @seealso \code{\link{orderly_pull_dependencies}} which provides a
+##'   higher-level interface to pulling from a remote (including
+##'   adding the downloaded archive into your orderly repository), and
+##'   see the documentation underlying the orderly remote driver that
+##'   your \code{orderly_config.yml} declares for information about
+##'   using the that remote.
```
weshinsley, line 296: Hah, a typo!!!

PR #210, opened 2020-04-15
VIMC-3659: Add the elapsed time into the database

richfitz:
This field has ~always~ almost always been present in the metadata but for some reason has never made it out into the database.  This is useful to know before rerunning reports, so let's write it into the database.  This will be required to reflect this information through OrderlyWeb

---

Edit: turns out that we introduced this about 0.5.0, which was ages ago (6940fd7f8f355b20ebcb81f35727ae785e89e917) but this means that a migration is required for the earliest reports. @r-ash - this will be the first migration that you've seen so let me know if it's not clear enough


inst/migrate/1.1.25.R
```diff
@@ -0,0 +1,12 @@
+migrate <- function(data, path, config) {
```
r-ash, line 1: Do you need some test of this migration? It looks like there is in `test-migrate.R` or most of the previous migrations.

richfitz: Good point, I'll add one

PR #212, opened 2020-04-16
VIMC-3640: Add support for posting back to teams

r-ash:
This PR will
* Add support for specifying `teams_url` in config
* Post success notification to teams if configured

Couple of things to discuss here
* This has no update of how the cfg is managed with `slack_url` or `teams_url` (I see there is a note about tidying up the `slack_url` handling. It is just using the slack code but generating a slightly different structure for the post content. Perhaps not the most extensible pattern if we see the notifications going somewhere else. This could be good opportunity for making something more extensible - we could use some event & listener like architecture to throw a `orderly_commit` event with all the meta information on there and then have multiple listeners (which other people could extend in plugins) which could respond to that. It would be pretty cool but perhaps a bit overkill for what we need here
* Have a look at the image in `orderly-feed` channel in teams - we probably want to use something else here. And I need to fix the profile image for the connector. Any thoughts on what you want those images to be?
* This uses `glue`, I can update to remove that dependency if you would rather not bring in another package


R/slack.R
```diff
@@ -1,62 +1,52 @@
 ## Send slack messages!
 
-slack_post_success <- function(dat, config) {
+post_success <- function(dat, config) {
```
richfitz, line 3: I think this a good way of handling this for now - posts to both rather than either, and we minimal extra debt.  When I do this internal structure refactor we'll have defined "phases" of execution and that will include something "hookable" - we can then move this stuff out there somehow perhaps.  But very open to other ideas of how to sort that out

R/slack.R
```diff
@@ -66,8 +56,76 @@ slack_data <- function(dat, remote_name, report_url, remote_is_primary) {
...
+  list(
+    "@type" = "MessageCard",
+    "@context" = "http://schema.org/extensions",
+    summary = content$fallback,
+    themeColor = col,
+    sections = list(
+      list(
+        activityTitle = content$title,
+        activityText = content$text,
+        ## TODO: Choose another image from somewhere?
```
richfitz, line 77: In the orderly feed on slack we use an ambulance? There must be some sort of emoji-as-a-service we could use here?

r-ash: > There's also a failing test that I think you've seen, but just looking over the implementation this seems like a good way to do it given the current constraints on the package. Having two things now will make it easier to get the refactor right I think.
> 
> I don't think we should try for anything too fancy at this point.
> 
> You should bump the version and add a news bullet before too, and if you can remove glue that'd be great.
> 
> Is there a Teams channel with an example you've used in testing? Could you post a a screenshot?

Thanks, fixed that test. Tested this posting to orderly-feed in vimc team. It looks like below atm.

Selection_046

I think the connector icon looks pretty terrible but sadly not many options to make this better from teams. You just upload an image and it puts it in a hexagon.

Let me know what you think.

R/slack.R
```diff
@@ -66,8 +56,76 @@ slack_data <- function(dat, remote_name, report_url, remote_is_primary) {
...
+  id <- dat$meta$id
+  name <- dat$meta$name
+  elapsed <- format(as.difftime(dat$meta$elapsed, units = "secs"), digits = 2)
+  git <- dat$git
+  if (!is.null(git)) {
+    branch <- git$branch %||% "(detached)"
+    sha <- git$sha_short
+    if (!is.null(git$github_url)) {
+      url <- sprintf("%s/tree/%s", git$github_url, sha)
+      sha <- glue::glue(link_format, url = url, label = sha)
```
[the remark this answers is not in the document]
r-ash: Removed that dependency and have moved docopt and crayon to suggests section too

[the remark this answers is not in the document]
richfitz: Sorry, wasn't meaning for you to have to do that, and it will require some fiddling to wrappers to make sure that the functionality works without them present - including some faff with docker perhaps. Could you move crayon and docopt back and we'll take care of that on VIMC-3664

R/slack.R
richfitz: I think this is great

r-ash:
> Could you just revert [82121bc](https://github.com/vimc/orderly/commit/82121bc030d0fd18bbee0fa4457b06904b528aa0) and then we can merge this (sorry I'd posted on a comment but it's probably lost in your emails)

Done :)

PR #213, opened 2020-04-18
orderly_envir.yml vignette example needs : instead of =

jeffeaton:
In the Getting Started vignette, I believe that the example for setting environment variable in `orderly_envir.yml`:

> `DATA_URL=https://www.example.com/thedata` 

needs to be yaml formatted:

> `DATA_URL: https://www.example.com/thedata`

When I used `DATA_URL=https://www.example.com/thedata` , I got the error message: `Error: 'orderly_envir.yml' must be named`. It worked as expected when I changed `=` to `: `.


richfitz:
Thanks @jeffeaton! @sangeetabhatia03 also spotted this and fixed in https://github.com/vimc/orderly/pull/215 🙈

PR #222, opened 2020-04-27
vimc-3702: Backup before rebuild

richfitz:
This would have been useful with the published problem as reported by Kim, but might be useful generally going forward (and the cost is low).  Uses the already-tested backup db that uses the "proper" way of backing up a SQLite database.


richfitz:
yes, in `/orderly/backup/db` on montagu, so that is then *actually* backed up by bb8 etc.  We do this already for destructive changes to the data migrations

PR #231, opened 2020-05-11
Add orderly workflows and yml reading

r-ash:
This PR will

* Add skeleton `orderly_workflow` function
* Adds yml reading

Doesn't actually do any report running but locates a yml file, reads it and parses this into a list of `steps`. 

Going to add some validation then this is ready to review.


R/workflow.R
```diff
@@ -0,0 +1,76 @@
...
+    }
+  )
+)
+
+parse_steps <- function(yml) {
+  unname(unlist(yml$steps))
+}
+
+validate_steps <- function(steps, config, workflow_name) {
+  reports <- basename(list_dirs(path_src(config$root)))
```
r-ash, line 68: This is the sample impl as `orderly_list` but using the config object. Looks like the impl of the `orderly_list` stuff if half way through a move to a `list2` function so wanted to mention this. Don't know whether it is worth adding an `orderly_list(config)` function which is called both here and in `orderly_list(root, locate)`? Or some other way of tidying up the code re-use

richfitz: yeah, there's a bunch of stuff like this that needs tidying up. I'm happy for you to tidy up here or in another ticket. It's internal stuff only so we can do it however we want

richfitz: Or you can just use `orderly_list(config)` which works fine as `orderly_config_get(config)` just returns `config`

R/workflow.R
```diff
@@ -0,0 +1,76 @@
...
+    initialize = function(workflow_name, workflow_path, config) {
+      self$workflow_name <- workflow_name
+      self$workflow_id <- ids::random_id()
+      self$config <- config
+      raw <- yaml_read(workflow_path)
+      self$steps <- parse_steps(raw)
+      validate_steps(self$steps, self$config, workflow_name)
+    },
+
+    run = function() {
```
richfitz, line 55: Consider making this just error while we don't have an implementation?

R/workflow.R
```diff
@@ -0,0 +1,76 @@
...
+      ids::random_id()
+    }
+  )
+)
+
+parse_steps <- function(yml) {
+  unname(unlist(yml$steps))
+}
+
+validate_steps <- function(steps, config, workflow_name) {
```
richfitz, line 67: I think this should use `check_fields` to validate that each step has just the configuration options we support

R/workflow.R
```diff
@@ -0,0 +1,76 @@
...
+  reports <- basename(list_dirs(path_src(config$root)))
+  for (step in steps) {
```
richfitz, line 69: perhaps report all back at once?  Elsewhere we use

```
msg <- setdiff(step, reports)
if (length(msg)) {
  stop(sprintf("Cannot run workflow '%s' as report missing: %s",
               paste(squote(msg), collapse = ", ")))
}
```

R/workflow.R
r-ash: Oh nice - I missed that. `orderly_list(config)` is perfect.

R/workflow.R
r-ash: Nice, added that now too.

PR #232, opened 2020-05-13
VIMC-3611: refactor orderly run

richfitz:
This is a large structural change that will likely take a few PRs to get completely right. I'd like to get us broadly there on this PR and then spin out minor bits to do into others. Additional things introduced in this commit that exceeded the scope I was looking for:

* `orderly::orderly_data` has been unceremoniously removed from the package - I don't think this was ever used and it's unlikely to be important to external users as it requires a SQL database to do anything useful
* `orderly::orderly_run` drops a bunch of arguments that are not very user-facing - this is really VIMC-3539/VIMC-3539 and we can walk it back for now if you think it's worthwhile

My aim here is not to come up with something perfect, but rationalise logic that was spread all over the show into something vaguely sensible; I can see a bunch of work we can do on this as we develop plugins (like notifiers and additional data sources).  There's quite a bit of code in orderly_run that could move into orderly_version and get renamed (e.g., info becomes version) but that can be done later and makes this PR harder to read


tests/testthat/test-run.R
```diff
@@ -1,79 +1,42 @@
 context("run")
 
-## Same as in read; we generate a report and then break it
 test_that("minimal", {
   path <- prepare_orderly_example("minimal")
   on.exit(unlink(path, recursive = TRUE))
 
-  config <- orderly_config$new(path)
```
richfitz, line 8: This test was a hot mess so I've just got rid of it and replaced with a sensible minimal test

richfitz:
> I think I'd be tempted to move the functions in orderly_version which only get called by the class into private ones but no big issue there.

I went back and forth on this, and ended up going public only until we decided what the interface should look like. We should do a round of proper encapsulation and interface design at some point perhaps?

> I also think adding some docs to the functions which are less clear about what they do or why they do it would be useful e.g. set_current not clear from the name what it achieves.

I'll have a quick look at this (and perhaps also do a mass private move) after lunch

PR #233, opened 2020-05-13
VIMC-3834: Add support for running a workflow

r-ash:
I think some room for improvement of the messages here. Including returning things like timings we do in orderly run and setting a colour for workflow messages. Happy to do this here or in another ticket.

I tried adding a `verbose` flag and have messages from `orderly_run` not printed by default but looks like no current way in `orderly_run` to do that so I dropped that arg for now. Probably good to leave that for another ticket.


R/workflow.R
```diff
@@ -14,13 +17,14 @@
...
+orderly_workflow <- function(name, envir = NULL, root = NULL, locate = TRUE,
+                             message = NULL, instance = NULL, remote = NULL) {
   ## Locate file
   config <- orderly_config_get(root, locate)
   workflow <- orderly_workflow_get(name, config)
 
-  workflow$run()
+  workflow$run(envir, message, instance, remote)
 
   ## TODO: Write to DB after run
```
richfitz, line 29: I think the DB work will follow pretty easily after #232 and we can (and should) do this one by one as we go.

R/workflow.R
```diff
@@ -50,10 +54,23 @@ workflow <- R6::R6Class(
...
+      run_ids <- unlist(lapply(output, function(out) {
+        out$value
+      }))
+      out_message <- sprintf("Completed running workflow '%s' with ID '%s'\n",
+                             self$workflow_name, self$workflow_id)
+      no_failed <- length(self$steps) - length(run_ids)
+      if (no_failed > 0) {
+        out_message <- sprintf("%s with %s failure(s)", out_message, no_failed)
+      }
+      orderly_log("workflow", out_message)
```
richfitz, line 72: If you look at `orderly_log_style` then you can get a custom colour on the workflow, which I think would be useful

R/workflow.R
```diff
@@ -76,3 +93,25 @@ validate_workflow <- function(raw_yml, config, workflow_name, workflow_path) {
   }
   steps
 }
+
+run_single <- function(report, envir, root, message, instance, remote) {
```
richfitz, line 97: workflow_run_single as a name perhaps?

R/workflow.R
```diff
@@ -14,13 +17,14 @@
 ##'
 ##' # To run most reports, provide the report name (and the path if
 ##' # not running in the working directory, as is the case here):
-##' #id <- orderly::orderly_workflow("my_workflow", root = path)
-orderly_workflow <- function(name, root = NULL, locate = TRUE) {
+##' ids <- orderly::orderly_workflow("my_workflow", root = path)
+orderly_workflow <- function(name, envir = NULL, root = NULL, locate = TRUE,
```
richfitz, line 21: We should consider if we support a "draft" mode or not here. In draft mode we'd not commit and we'd pass in `use_draft = "newer"` to orderly run; that might be better for local use, but then it's not clear how often we'd want local use of this? Could push this to another ticket quite happily

R/workflow.R
```diff
@@ -50,10 +54,23 @@ workflow <- R6::R6Class(
...
+                                      self$workflow_name, self$workflow_id))
+      output <- lapply(self$steps, run_single, envir, self$config, message,
+                       instance, remote)
+      run_ids <- unlist(lapply(output, function(out) {
+        out$value
+      }))
+      out_message <- sprintf("Completed running workflow '%s' with ID '%s'\n",
+                             self$workflow_name, self$workflow_id)
+      no_failed <- length(self$steps) - length(run_ids)
+      if (no_failed > 0) {
```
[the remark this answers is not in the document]
richfitz: Also, behaviour on failure needs dealing with: we need to consider if we can "uncommit" a report (see VIMC-3844) as there are are a few corner cases.  So perhaps for now we leave them in and note that this is not atomic and think about this more. Same problem affects batch

R/workflow.R
richfitz: You might also want to do this last one as a multiline message (if the 2nd arg to `orderly_log` is a character vector you get nice printing over multiple lines) with `<name>:<id>` pairs?

richfitz: When we add cli support for this we should wire up the notifications too

R/workflow.R
```diff
@@ -50,10 +54,23 @@ workflow <- R6::R6Class(
...
-      ## TODO: Implement run
-      orderly_log("workflow", paste0("Running workflow ", self$workflow_id))
-      stop("Workflow running not yet implemented")
+    run = function(envir = NULL, message = NULL, instance = NULL,
+                   remote = NULL) {
+      orderly_log("workflow", sprintf("Running workflow '%s' with ID '%s'",
+                                      self$workflow_name, self$workflow_id))
+      output <- lapply(self$steps, run_single, envir, self$config, message,
+                       instance, remote)
+      run_ids <- unlist(lapply(output, function(out) {
```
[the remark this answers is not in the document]
r-ash: That's way better, thanks

R/workflow.R
r-ash: Stuck that in another task https://mrc-ide.myjetbrains.com/youtrack/issue/VIMC-3845

R/workflow.R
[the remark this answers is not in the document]
r-ash: Yeah we could move to run them all and then only commit if all manage to run successfully. Or have some interactive message "x failed y passed, do you want to commit?" if some things fail. I think lots of things to do. 

I didn't make the workflow fail here as it feels weird to me that if you return a failure form the workflow run when some reports could have successfully completed. If it fails how do you get the id of the reports which succeeded? But definitely current behaviour isn't great. 

I will add a ticket to make this better and I can make this fail for the time being if any of the reports fail?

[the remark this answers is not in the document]
r-ash: https://mrc-ide.myjetbrains.com/youtrack/issue/VIMC-3847

R/workflow.R
r-ash: Made these print green now. The style is just "workflow" which isn't very generic unlike the others (highlight, alert) so happy to change that if you would prefer something more general?

PR #236, opened 2020-05-26
Capture error in failed run

richfitz:
This PR allows the error thrown in a failed orderly run to be recorded in the logfile - currently it is only printed to screen.  This has proved quite annoying in parallel runs. We also print the traceback here too


docker/build_website
```diff
@@ -9,8 +9,8 @@ HERE=$(dirname $0)
 ORDERLY_BASE="$TAG_SHA"
 ORDERLY_DEV="${PACKAGE_ORG}/${PACKAGE_NAME}-dev:${GIT_SHA}"
 
-USER_UID=`id -u`
-USER_GID=`id -g`
+USER_UID=$(id -u)
```
richfitz, line 12: This is unrelated, but opportunistic to fix two remaining codefactor complaints (no functional change)

PR #237, opened 2020-05-27
VIMC-3873: Add workflow to CLI

r-ash:
This PR will
* Expose `orderly workflow` to CLI tool
* Changes return from `orderly_workflow` to return ID and name of each report run

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,7 @@
 # orderly 1.2.5
 
 * `orderly::orderly_workflow` now returns name and ID of reports run
-* Add ability to run workflow from CLI
+* Add ability to run workflow from CLI (VIMC-3873)
 
 # orderly 1.2.4
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,5 @@
 # orderly 1.2.5
 
-* `orderly::orderly_workflow` now returns name and ID of reports run
 * Add ability to run workflow from CLI (VIMC-3873)
 
 # orderly 1.2.4
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,8 @@
+# orderly 1.2.5
+
+* `orderly::orderly_workflow` now returns name and ID of reports run
+* Add ability to run workflow from CLI
+
 # orderly 1.2.4
 
 * Error messages and stack traces are now preserved in `orderly.log` after a faied run; this primarily impacts the cli runner and then primarily when being run in parallel where the log is not printed to screen (VIMC-3841)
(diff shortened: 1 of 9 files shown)


R/main.R
```diff
@@ -393,6 +398,57 @@ main_do_batch <- function(x) {
...
+  if (pull) {
+    git_pull(config$root)
+  }
+
+  run_reports <- orderly_workflow(name, root = config, instance = instance,
+                                  message = message)
+  output <- vcapply(run_reports, function(report) {
+    path_rds <- path_orderly_run_rds(
+      file.path(config$root, "archive", report$name, report$id))
+    post_success(readRDS(path_rds), config)
```
r-ash, line 445: Because we post the success here which needs the report name and ID I modified `orderly_workflow` to return both the id and the name of the reports it has run. That feels like a fine change to me but is now inconsistent with batch & run but as they only ever run for 1 report seems fine.

Alternative is to post something different for a workflow, i.e. post once the workflow has completed that is has run and includes reports x, y, z with ids x, y, z. I can always do that in another PR.

richfitz: I like the idea of it posting as it goes because then you do get more progress information - can we move this into the `workflow$run` method perhaps?  This feels similar to the internal- and external- facing view on orderly run that we need to think about properly at some point.

Posting something afterwards for the workflow might be nice, and we could also tweak the `post_success` to indicate that we're part of a workflow. Another thing we might want to add to orderly web is a workflow page - so we could link there. Lots of things to add! But of this set of things need doing here.

NEWS.md
```diff
@@ -1,3 +1,8 @@
+# orderly 1.2.5
+
+* `orderly::orderly_workflow` now returns name and ID of reports run
```
richfitz, line 3: Can you put the ticket here please?

R/main.R
```diff
@@ -393,6 +398,57 @@ main_do_batch <- function(x) {
...
+  ## * error while running report
+  ## * error checking artefacts
+  config <- orderly_config_get(x$options$root, TRUE)
+  name <- x$options$name
+  instance <- x$options$instance
+  print_log <- x$options$print_log
+  pull <- x$options$pull
+  message <- x$options$message
+
+  if (print_log) {
```
richfitz, line 429: We really need to deal with this at some point...

R/main.R
r-ash: Having a look this already is being done in `workflow_version` on a `commit` so I have added a test which checks that `post_success` gets called for each report in a workflow. Can add a ticket for expanding `post_success` to have some indication that we are part of a workflow

PR #238, opened 2020-05-27
VIMC-3854: Add workflows vignette

r-ash:
I might ask one of the science team to take a look at this too to see if it is good enough to understand.

Diff is noisy here, I think easier to look at the two vignettes separately. I have changed some of wording in previous `workflows` vignette as suggested. But there are still usages of that in the text.


vignettes/workflows.Rmd
```diff
@@ -1,208 +1,67 @@
...
-      token: $GITHUB_TOKEN
+``` {r}
+out <- orderly::orderly_workflow("my_workflow", root = path)
 ```
 
-Now, as part of the pull request review procedure, the reviewer will run the report on the staging environment, which means that it should work without error on the production environment if they have been set up the same way.
-
-```r
-orderly::orderly_run_remote("report-a", remote = "staging", ref = "origin/myfeature")
+This returns a list containing the name and ID of each of the reports run as part of the workflow
```
r-ash, line 64: This isn't true at the moment but will be if  #237 is merged in without any changes

vignettes/workflows.Rmd
```diff
@@ -1,208 +1,67 @@
...
-* On the central server, the archive directory is an "append only" system; nothing is ever modified.  This seems wasteful at first as there can be lots of copies of a report, but disk space is cheaper than researcher time (and you can remove redundancy with `orderly::orderly_deduplicate`)
-
-## Group of researchers, sharing via server
-
-This is the model that we use within the [MRC Centre for Global Infectious Disease Analysis](https://www.imperial.ac.uk/mrc-global-infectious-disease-analysis), and it does require considerably more setup, but also more reliability and flexibility.
-
-As above, we use a single "source tree", kept on GitHub or similar.  However, someone in the group takes responsibility for setting up a copy of [OrderlyWeb](https://github.com/vimc/orderly-web), our open source, self-hosted orderly server.  The OrderlyWeb server software provides a number of things
+Orderly supports running workflows - these are pre-configured lists of reports which are run in sequence as part of an analysis.
```
richfitz, line 39: I think a bit more motivation here would be useful - something the science team might want to add? But basically "Suppose you had a sequence of reports that you always want to run..." show the problem of doing this manually perhaps

vignettes/patterns.Rmd
```diff
@@ -0,0 +1,208 @@
+---
+title: "Work patterns with orderly"
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{Work patterns with orderly}
```
richfitz, line 5: What about simply "orderly patterns" as the title?

vignettes/patterns.Rmd
```diff
@@ -0,0 +1,208 @@
...
+```
+
+This is especially useful if `report-b` has many dependencies, or depends on particular versions of a report.
+
+### Some comments on collaboration
+
+There are a number of features about the system here that might not be obvious at first, but that will aid collaboration.
+
+* Because each report exists in its own directory within `src/` we reduce the potential for conflicts.  In a collaborative analysis where multiple files are shared across analyses, it is easy to have conflicts.  `orderly` forces analyses into separate directories, which means most of the time it is trivial to merge even long-lived branches.  This is particularly important with relatively new users of git.
+* Because every report gets a unique identifier, it is straightforward to combine outputs that were run on different machines.  In the example above this is a centralised system (one canonical source, several destinations) but in theory more complicated distributed workflows are possible.
```
richfitz, line 104: workflows can be patterns here I think

vignettes/patterns.Rmd
```diff
@@ -0,0 +1,208 @@
...
+With some additional development on `orderly`, further work patterns are possible, including a decentralised approach.  We might like to have a set of known "good" archives, or partially overlapping source trees, or archives that depend on data sources that come from other orderly archives that are not even present in their source tree.
+
+To support this, we can have a network of orderly servers which can exchange sub-trees amongst each other.  In this fashion, we can imagine the following things being possible
+
+* Have common data processing tasks described in one orderly source tree, but used by several other sources trees
+* Have orderly "runners" available on HPC platforms, making it easy to reliably run demanding tasks on larger infrastructure then incorporate them back into the "canonical" archive
+* Allow distributed teams of researchers with intermittent access to synchronise their own trees of outputs, eventually becoming consistent with each other
+
+All the lower-level support for these workflows is available, and implementing them in orderly is not necessarily hard - we just have not found the need for this ourselves yet.  If this is of interest please get in touch.
```
richfitz, line 208: workflows here can be patterns

vignettes/workflows.Rmd
r-ash: Tidied that up, asked Susy to take a look too

r-ash:
> Really like the workflows description. I guess the vignette will be updated once new arguments are included in the workflows function (such as the remote and instance argument). Also, it maybe be good to include a small description about which working directory we should run the function. It is on the `./src ` ` ./workflows` or the root of montagu-report i.e.,` ./` It is still not clear to me

Thanks Susy, yeah we will update when further functionality is added to the workflows. Good news is `remote` and `instance` are supported already. I've added a note about that in the workflow. I've included a bit which explains the working directory situation too - thanks for that.

PR #240, opened 2020-06-02
VIMC-3857: Orderly support for git branches endpoint

r-ash:
This PR will

* Add a function to orderly_runner to return the unmerged git branches

Can optionally include `master` branch which I think is what we want to do in `orderly.server` and also ignores a `gh-pages` branch. 

I expect the format of the date may change here depending on what Emma wants in the front end so we should probably only merge this once corresponding `orderly.server` has been checked.


R/git.R
```diff
@@ -85,6 +85,22 @@ git_fetch <- function(root = NULL) {
...
 git_pull <- function(root = NULL) {
-  orderly_log("git", "fetch")
+  orderly_log("git", "pull")
   git_run("pull", root = root, check = TRUE)
 }
+
+git_branches_no_merged <- function(root = NULL, include_master = FALSE) {
+  branches <- git_run(c("for-each-ref", "--sort=-committerdate",
+                        "--format='%(refname:short),%(committerdate)'",
+                        "--no-merged=master"), root = root, check = TRUE)$output
```
richfitz, line 95: How do we deal with old stale branches? There must be a bunch of dead ends in montagu-reports that we'd not really want to show? But it's not the end of the world either?

R/git.R
```diff
@@ -85,6 +85,22 @@ git_fetch <- function(root = NULL) {
 }
 
 git_pull <- function(root = NULL) {
-  orderly_log("git", "fetch")
+  orderly_log("git", "pull")
   git_run("pull", root = root, check = TRUE)
 }
+
+git_branches_no_merged <- function(root = NULL, include_master = FALSE) {
```
richfitz, line 92: When do we not want to include master?

r-ash: Never I don't think. Put this in as having `git_branches_no_merged` return master is different to how git behaves with `--no-merged` arg. So putting it as an arg was just to make it explicit. It also works slightly differently to the other branches as it is always first in the list. Happy to remove that though and return the branches we want.

R/git.R
r-ash: Stale branches would be shown at the moment. In some post processing we could just exclude branches which `last_commit` is older than a certain time. What would be appropriate? A couple of months?

richfitz: Let's go with everything for now, and later we can process that via the runner

PR #241, opened 2020-06-04
VIMC-3941: Add orderly support for returning git commit info

r-ash:
This PR will

* Add a function `git_commits` to orderly_runner which takes parameter branch which if master returns the last 25 commits, otherwise returns the commits not merged into master (limit 25)

Do we want to limit to return last 25 commits or limit based on some time period instead?

### DIFF
--- a/R/git.R
+++ b/R/git.R
@@ -112,10 +112,16 @@ git_branches_no_merged <- function(root = NULL, include_master = FALSE) {
 ## This gets last 25 commits from master
 ## if not master then gets the unmerged commits (limit 25)
 git_commits <- function(branch, root = NULL) {
-  commits <- git_run(c("log", "--pretty='%h,%cd'",
-                       "--date=unix", "--max-count=25",
-                       sprintf("refs/remotes/origin/%s", branch)),
-                     root = root, check = TRUE)$output
+  if (branch == "master") {
+    args <- c("log", "--pretty='%h,%cd'", "--date=unix", "--max-count=25",
+              sprintf("refs/remotes/origin/%s", branch))
+  } else {
+    remote_branch <- sprintf("refs/remotes/origin/%s", branch)
+    args <- c("log", "--pretty='%h,%cd'", "--date=unix", "--max-count=25",
+              sprintf("--cherry refs/remotes/origin/master...", remote_branch),
+              remote_branch)
+  }
+  commits <- git_run(args, root = root, check = TRUE)$output
   commits <- utils::read.table(text = commits, stringsAsFactors = FALSE,
                                sep = ",", col.names = c("id", "date_time"))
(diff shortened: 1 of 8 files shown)


R/git.R
```diff
@@ -101,12 +101,41 @@ git_branches_no_merged <- function(root = NULL, include_master = FALSE) {
...
+
+## This gets last 25 commits from master
+## if not master then gets the unmerged commits (limit 25)
+git_commits <- function(branch, root = NULL) {
+  if (branch == "master") {
+    args <- c("log", "--pretty='%h,%cd'", "--date=unix", "--max-count=25",
+              sprintf("refs/remotes/origin/%s", branch))
+  } else {
+    remote_branch <- sprintf("refs/remotes/origin/%s", branch)
+    args <- c("log", "--pretty='%h,%cd'", "--date=unix", "--max-count=25",
```
richfitz, line 120: nice - just checked and this seems to get exactly the right sort of thing on unmerged branches that have diverged from master

R/git.R
```diff
@@ -101,12 +101,41 @@ git_branches_no_merged <- function(root = NULL, include_master = FALSE) {
...
+              sprintf("--cherry refs/remotes/origin/master...", remote_branch),
+              remote_branch)
+  }
+  commits <- git_run(args, root = root, check = TRUE)$output
+  commits <- utils::read.table(text = commits, stringsAsFactors = FALSE,
+                               sep = ",", col.names = c("id", "date_time"))
+  commits$age <- calculate_age(commits$date_time)
+  commits$date_time <- convert_unix_to_iso_time(commits$date_time)
+  ## ID can be parsed as an integer by read.table if by chance the id contains
```
richfitz, line 129: So good that you hit this now and not later...

R/git.R
```diff
@@ -101,12 +101,41 @@ git_branches_no_merged <- function(root = NULL, include_master = FALSE) {
...
+  ## only numbers
+  commits$id <- as.character(commits$id)
+  commits
+}
+
+calculate_age <- function(times) {
+  rep(as.integer(Sys.time()), length(times)) - times
```
richfitz, line 136: This needs a unit on it:

```
now <- Sys.time()
as.integer(now - (now - 10000)) # 2 (because close to 2 hours)
as.integer(as.numeric(now - (now - 10000), "secs")) # 10000
```

(Could you add a unit test too when fixing?)

R/git.R
r-ash: Yeah - very fortunate I got this in one of runs of the test!

R/git.R
r-ash: I think the previous impl avoids this as we cast to int before doing the calculation so I was having trouble recreating this in a test. I've updated the impl anyway as better to be explicit about the conversion and added a test.

PR #242, opened 2020-06-04
VIMC-3945: Add orderly support for listing available reports for a particular git branch and commit

r-ash:
Couple of things to note here this follows on from #241

This fixes a bug in #241 and adds a test for it. I've left that in here as the test which recreates it is wrapped up with some testing of the feature added here. But I can separate out if you prefer.

### DIFF
--- a/R/git.R
+++ b/R/git.R
@@ -118,7 +118,7 @@ git_commits <- function(branch, root = NULL) {
   } else {
     remote_branch <- sprintf("refs/remotes/origin/%s", branch)
     args <- c("log", "--pretty='%h,%cd'", "--date=unix", "--max-count=25",
-              sprintf("--cherry refs/remotes/origin/master...", remote_branch),
+              sprintf("--cherry refs/remotes/origin/master..%s", remote_branch),
               remote_branch)
   }
   commits <- git_run(args, root = root, check = TRUE)$output
--- a/R/git.R
+++ b/R/git.R
@@ -135,13 +135,22 @@ git_commits <- function(branch, root = NULL) {
 get_reports <- function(branch, commit, root) {
   if (branch == "master") {
     ## Get all reports in commit if on master branch
-    args <- c("ls-tree", "--name-only", "-d", sprintf("%s:src/", commit))
+    reports <- git_run(c("ls-tree", "--name-only", "-d",
+                         sprintf("%s:src/", commit)),
+                       root = root, check = TRUE)$output
   } else {
-    ## Get only files which have changed from master copy
-    ## Note this could inclue files as well as directories.
-    ## How can we exclude them?
-    args <- c("diff-tree", "--name-only",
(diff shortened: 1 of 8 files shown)


R/git.R
```diff
@@ -101,12 +101,55 @@ git_branches_no_merged <- function(root = NULL, include_master = FALSE) {
...
+}
+
+get_reports <- function(branch, commit, root) {
+  if (branch == "master") {
+    ## Get all reports in commit if on master branch
+    args <- c("ls-tree", "--name-only", "-d", sprintf("%s:src/", commit))
+  } else {
+    ## Get only files which have changed from master copy
+    ## Note this could inclue files as well as directories.
+    ## How can we exclude them?
```
r-ash, line 150: Also wanted to discuss this point. 

So I tried checking those which are directories using R but that can't be done if it doesn't exist locally (which can definitely happen here). I can't see an obvious flag to `diff-tree` like there is on `ls-tree` for only returning directories. And similarly piping to `du` to only get dir doesn't work for same reasons as checking for dirs in R doesn't work.

I could do something like get return from `diff-tree` return all changed files and then pipe it to `dirname` to only give the dirname part we are interested in.

richfitz: Git only stores files (not directories) which means if you pipe it to `dirname()` then get the *first* element below `src` you your full list of directories that you are interested in - as you suggest above.

R/git.R
```diff
@@ -139,3 +139,17 @@ calculate_age <- function(times) {
...
+    args <- c("diff-tree", "--name-only",
+              sprintf("refs/remotes/origin/master:src/..%s:src/", commit))
```
richfitz, line 152: For unmerged branches that can't be fast-forwarded to master I think this is incorrect as I am seeing a lot of reports changed:

```
git diff-tree --name-only refs/remotes/origin/master:src/..origin/vimc-3943:src
```

gives 1 result as wanted

but

```
git diff-tree --name-only refs/remotes/origin/master:src/..origin/vimc-3347_2:src
```

gives 111!

I see better behaviour with

```
git diff --name-only refs/remotes/origin/master...origin/vimc-3347_2 -- src
```

which we can get the relevant things with:

```
git diff --name-only refs/remotes/origin/master...origin/vimc-3347_2 -- src | xargs -I {} dirname {} | sed -E 's;src/([^/]*).*;\1;' | uniq
```

But this is much better done in R with the list from git

richfitz: Not sure what a great test would be here - if you have the stomach for it a little git demo that has divergent branches would be very useful I think

R/git.R
r-ash: Thanks, fixed this in R side now

R/git.R
r-ash: Nice one, thanks. Have got a test which recreates that and a fix in now.

PR #243, opened 2020-06-08
VIMC-3953: Add function to get parameters for a report at a commit id

r-ash:
Diff here noisy again as this builds on #242

Plan here is to return as a parsed list without any further processing. Then do the conversion from this list representation into the format we want to return from the endpoint in orderly.server

One thing to note is that I see there is an old format for the parameters
```
parameters:
  - param1
  - param2
```

Which is deprecated but still supported - there is some conversion code around. This currently doesn't support that but if we want to then I should probably do the conversion from the migration step here before returning it to orderly.server.

### DIFF
--- a/tests/testthat/test-git.R
+++ b/tests/testthat/test-git.R
@@ -379,3 +379,45 @@ test_that("get_reports only shows one sided changes", {
   other_reports <- get_reports("other", other_commits$id, path[["local"]])
   expect_equal(other_reports, "other")
 })
+
+test_that("can get parameters from a report", {
+  testthat::skip_on_cran()
+  path <- prepare_orderly_git_example()
+
+  ## Write some parameters lines to the upstream yml and commit it
+  origin_yml <- file.path(path[["origin"]], "src", "minimal", "orderly.yml")
+  yml <- readLines(origin_yml)
+  text <- c(yml, c(
+    "parameters:",
+    "  a: ~",
+    "  b:",
+    "    default: test",
+    "  c:",
+    "    default: 2"
+  ))
+  writeLines(text, origin_yml)
+  invisible(git_run(c("add", "."), root = path[["origin"]]))
+  invisible(git_run(c("commit", "-m", "'add parameters'"),
+                    root = path[["origin"]]))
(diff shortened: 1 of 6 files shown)


r-ash:
> Been thinking about this a bit...
> 
> The migration code for parameters is very simple - `recipe_migrate` has:
> 
> ```r
>   if (is.character(raw$parameters)) {
>     msg <- c("Use of strings for parameters: is deprecated and will be",
>              "removed in a future orderly version - please use",
>              "'name: ~' for each parameter instead (or add a default).",
>              "See the main package vignette for details")
>     orderly_warning(flow_text(msg))
>     raw$parameters <- set_names(rep(list(NULL), length(raw$parameters)),
>                                 raw$parameters)
>   }
> ```
> 
> so the last bit of that could be pulled into a function and reused here safely I think, though that will be a bit tedious once we move the runner into orderly.server
> 
> A more general question is - what if the report on a branch is invalid? How do we handle that gracefully? At present, we'll return a 500 error eventually (depending on how we handle this in orderly.server).
> 
> So perhaps we just check that the parameters are in the new style and throw an error if we can't deal with it (orderly.yml can't be found, is not valid yml or if parameters are not in new format) and turn that into a 400 error in orderly.server

Thanks, improved the error handling here a little bit. In orderly capturing errors from call to git if either branch or orderly.yml can't be found. And also throwing an error from here if the yml can't be read. The code in `orderly.server` deals with if the structure of parameters not in an expected format.

tests/testthat/test-git.R
```diff
@@ -379,3 +379,73 @@ test_that("get_reports only shows one sided changes", {
...
+  expect_error(
+    get_report_parameters("rep1", "commit1", "."),
+    paste0(
+      "Failed to get report parameters for report 'rep1' and commit 'commit1':",
+      "\nNon zero exit code from git"))
+
+  mockery::stub(get_report_parameters, "git_run", list(
+    success = TRUE,
+    code = 0,
+    output = "[invalid_yml"
```
richfitz, line 444: Very invalid yml 😁

PR #245, opened 2020-06-11
VIMC-4048: Export orderly_config

r-ash:
This PR

* Updates `orderly_config` class to be called `orderly_config_`
* Renames `get_orderly_config` to `orderly_config` and exports it
* Locks bindings on `orderly_config_` for all fields except for `root` and `archive_version` which do get changed sometimes, and moved `migrate` and `validate` to private functions
* Adds roxygen docs for class

Particularly feedback on the cfg class documentation would be good, mostly I have tried to piece this together from other docs and the code.


NAMESPACE
```diff
@@ -9,6 +9,8 @@ export(orderly_bundle_pack)
 export(orderly_bundle_run)
 export(orderly_cleanup)
 export(orderly_commit)
+export(orderly_config)
+export(orderly_config_)
```
richfitz, line 13: I don't think we should export the class itself - I know that means documenting fairly manually at present, but I think we're not necessarily at the point where we can stand behind the API here either.

R/config.R
```diff
@@ -180,6 +248,15 @@ config_validate <- function(self, filename) {
 
   self$archive_version <- read_orderly_archive_version(".")
 
+  lockBinding(quote(destination), self)
```
richfitz, line 251: We already have one bit of locking in the initialize function, and I think this would go more naturally there.

Perhaps do this with a utility function that either locks all data members in a for loop or takes a vector of names and loops over that to lock them - we'll use this elsewhere

R/config.R
```diff
@@ -1,45 +1,93 @@
...
-
-
-orderly_config <- R6::R6Class(
+##' Orderly config
+##'
+##' An object representing orderly config. Reads the yml, validates
+##' it and makes the options available.
+##'
+##' @return An R6 object representing the orderly config.
+##' @export
```
richfitz, line 40: If you remove `@export` and add `@rdname orderly_config` we might get lucky and this will appear in the `orderly_config()` help without directly exposing the class

NEWS.md
```diff
@@ -1,3 +1,7 @@
+# orderly 1.2.12
+
+* Add new function `orderly_config` to return an object representing the config (VIMC-4048)
```
richfitz, line 3: I think we've used `orderly::orderly_` elsewhere in the NEWS

NAMESPACE
r-ash: Thanks, tidied that up now.

r-ash:
I've made the changes request in the review. But we saw a test failure where for some reason git commit hash was "Inf". I've not made any changes since seeing that but couldn't recreate it locally and on re-running the tests here they have passed. So possibly we have an intermittent test. Not sure what we can do here when I can't recreate it.

r-ash:
> I've made the changes request in the review. But we saw a test failure where for some reason git commit hash was "Inf". I've not made any changes since seeing that but couldn't recreate it locally and on re-running the tests here they have passed. So possibly we have an intermittent test. Not sure what we can do here when I can't recreate it.

Looks like this was from the duck typing `read.table` if it got a response from git like "123e456,123512387". It would cast the first to a numeric type beyond the range it represents and so show Inf.

R/config.R
```diff
@@ -1,45 +1,81 @@
...
+    stop("Reached root without finding 'orderly_config.yml'")
+  }
+  orderly_config_$new(root)
+}
 
-orderly_config <- R6::R6Class(
+##'
+##' @rdname orderly_config
+orderly_config_ <- R6::R6Class(
   "orderly_config",
```
richfitz, line 37: Can you add `cloneable = FALSE` to stop the noise in the docs about the clone method?

PR #246, opened 2020-06-16
VIMC-4073: Remove orderly runner

r-ash:


R/util.R
```diff
@@ -943,26 +943,6 @@ with_retry <- function(callback, n = 10, backoff = 1, match = NULL) {
                result$value$message), call. = FALSE)
 }
 
-calculate_age <- function(times) {
```
richfitz, line 946: protect and periodic can come out of here

PR #248, opened 2020-07-09
vimc-4131: Upgrade R

richfitz:
XML is not installing, causing other packages not to install, as it is the first thing to declare a dependency on 4.0.0 or more. I think that we're probably ok to upgrade now it's been out for a couple of months, though I think it breaks stan (we don't use that anywhere, and stan is a fragile thing to get running anyway unfortunately)

### DIFF
--- a/docker/Rprofile.site
+++ b/docker/Rprofile.site
@@ -1,2 +1,4 @@
-options(repos = c(CRAN = "https://cloud.r-project.org"),
-        download.file.method = "libcurl")
+options(repos = c(CRAN = 'https://packagemanager.rstudio.com/all/__linux__/focal/latest'), download.file.method = 'libcurl')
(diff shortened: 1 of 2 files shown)


docker/Rprofile.site
```diff
@@ -1,2 +1,4 @@
-options(repos = c(CRAN = "https://cloud.r-project.org"),
-        download.file.method = "libcurl")
+options(repos = c(CRAN = 'https://packagemanager.rstudio.com/all/__linux__/focal/latest'), download.file.method = 'libcurl')
```
richfitz, line 1: This is actually the same as what is present in the current version of the r-ver image but unless you track latest they fall back on a version pinned to the "last cran update before the release of the following version" which causes all sorts of problems

PR #250, opened 2020-07-30
VIMC-4193: Add support for running workflow with ref arg from CLI

r-ash:


R/workflow.R
```diff
@@ -71,13 +82,24 @@ workflow <- R6::R6Class(
         error = function(e) {
           orderly_log("error", sprintf("Running report '%s' failed", report))
           stop(e)
-        })
+        },
+        finally = git_restore())
```
richfitz, line 86: I think that this is too late, and you might use the git change in the main orderly run; there, we change git branch only during setup - see here: https://github.com/vimc/orderly/blob/d1628a254ac67e044422fbbfb84c6a71afbe067d/R/orderly_version.R#L393-L406

R/workflow.R
```diff
@@ -71,13 +82,24 @@ workflow <- R6::R6Class(
...
       }
       orderly_log("workflow",
                   sprintf("Completed running workflow '%s' with ID '%s'",
                           self$workflow_name, self$workflow_id))
       run_ids
     }
+  ),
+
+  private = list(
+    git_checkout = function(ref) {
```
richfitz, line 96: Just looking at this makes me think we really need to start thinking about https://mrc-ide.myjetbrains.com/youtrack/issue/VIMC-3900 at some point...

R/main.R
```diff
@@ -434,11 +436,16 @@ main_do_workflow <- function(x) {
   }
 
   if (pull) {
-    git_pull(config$root)
+    if (is.null(ref)) {
```
richfitz, line 439: As this duplicates this code (https://github.com/vimc/orderly/blob/d1628a254ac67e044422fbbfb84c6a71afbe067d/R/main.R#L185-L192) can you factor into its own function please?

R/workflow.R
r-ash: Thanks, fixed that now

PR #257, opened 2020-11-09
VIMC-4275: Return report id as progress message

r-ash:


r-ash:
Going to close this as we don't need it yet - might be used in future.

PR #262, opened 2021-01-08
Make orderly_pull_archive parameter arg more behaviour more predictable

r-ash:
This PR will:
* Construct a default query from parameters if `id = "latest"`
* Update docs to hopefully make this behaviour clear

### DIFF
--- a/R/query_search.R
+++ b/R/query_search.R
@@ -450,3 +450,26 @@ remote_report_metadata <- function(name, remote, config) {
   candidates$type <- "remote"
   candidates[c("id", "type", "path")]
 }
+
+
+build_query <- function(parameters) {
+  withCallingHandlers({
+    lapply(names(parameters), function(name) {
+      assert_scalar(parameters[[name]], name)
+    })
+  }, error = function(e) {
+    e$message <- paste0("Failed to build default query:\n", e$message)
+    stop(e)
+  })
+  quote_character <- function(x) {
+    out <- x
+    if (is.character(x)) {
+      out <- paste0('"', x, '"')
+    }
+    out
+  }
+  params <- lapply(parameters, quote_character)
+  sprintf("latest(%s)",
+          paste(sprintf("parameter:%s == %s", names(params), unlist(params)),
+                collapse = " && "))
+}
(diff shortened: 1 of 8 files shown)


R/query_search.R
```diff
@@ -450,3 +450,26 @@ remote_report_metadata <- function(name, remote, config) {
...
+    stop(e)
+  })
+  quote_character <- function(x) {
+    out <- x
+    if (is.character(x)) {
+      out <- paste0('"', x, '"')
+    }
+    out
+  }
+  params <- lapply(parameters, quote_character)
```
richfitz, line 471: I had suggested

```
  params <- sprintf(dquote(list_to_character(lapply(parameters, as.character)))
```

but realise that this is even easier:

```suggestion
  params <- sprintf("parameter:%s == %s", names(parameters), names(parameters))
```

and then just paste that together with ` && `.

The query interface should do the validation of the parameters for us, we should not do the conversion to character etc here

R/remote.R
```diff
@@ -49,13 +49,19 @@
...
 ##'   resolution.  If you are using a query for \code{id} that
 ##'   involves a parameter (e.g., \code{latest(parameter:x == p)}) you
 ##'   will need to pass in the parameters here.  Similarly, if you are
 ##'   pulling a report that uses query dependencies that reference
 ##'   parameters you need to pass them here (the same parameter set
-##'   will be passed through to all dependencies).
+##'   will be passed through to all dependencies). If passed with
+##'   \code{id = "latest"} then query will be constructed to pull latest
+##'   version of report with report parameters equal to those passed e.g. if
+##'   \code{parameters = list(nmin = 0.5, nmax = 1)} this will pull latest
```
richfitz, line 61: I'd write out here the implied query, something like

> this is equivalent to providing the query `latest(parameter:nmin = nmin && parameter:nmax:nmax)` along with your parameters

richfitz (changes requested):
Couple of things that will simplify this, sorry forgot how this all works when we chatted as it's actually even simpler than we'd thought

R/query_search.R
r-ash: ooh ofc yes

richfitz (approved):

PR #266, opened 2021-02-10
Check ref is valid before calling run remote

r-ash:
We will need this after orderly.server refactor is merged, but I think no problem in merging this first?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# orderly 1.2.22
+# orderly 1.2.23
 
 * `orderly_run_remote` checks that ref exists before running (VIMC-4574)
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# orderly 1.2.22
+
+* `orderly_run_remote` checks that ref exists before running (VIMC-4574)
+
 # orderly 1.2.20
 
 * `orderly_pull_archive` is now more tolerant of trailing slashes (#260)
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,10 @@
-# orderly 1.2.23
+# orderly 1.2.24
 
 * `orderly_run_remote` checks that ref exists before running (VIMC-4574)
+
+# orderly 1.2.23
+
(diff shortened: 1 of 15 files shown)


R/git.R
```diff
@@ -71,3 +71,18 @@ git_pull <- function(root = NULL) {
...
+git_ref_is_valid <- function(ref, root) {
+  if (is.null(ref)) {
+    return(invisible(TRUE))
+  }
+  tryCatch(
+    ## ref can be either a branch name or a git hash, git_commit_id will
+    ## give us full ID if ref is valid, if not throws an error.
+    gert::git_commit_id(ref, root),
+    error = function(e) {
+      stop(sprintf("Failed to locate git ref %s", ref))
```
richfitz, line 84: Is the default error from gert not good enough here?

R/remote.R
```diff
@@ -247,7 +247,9 @@ orderly_run_remote <- function(name, parameters = NULL, ref = NULL,
                                progress = TRUE,
                                root = NULL, locate = TRUE, instance = NULL,
                                remote = NULL) {
-  remote <- get_remote(remote, orderly_config(root, locate))
+  cfg <- orderly_config(root, locate)
+  git_ref_is_valid(ref, cfg$root)
```
richfitz, line 251: I think at this point we should translate ref from branch to hash, no? So that way we will run the commit that the researcher sees at some point in the future, even if more commits are pushed to that branch

NEWS.md
```diff
@@ -1,3 +1,7 @@
+# orderly 1.2.22
```
richfitz, line 1: Looks like the version number needs to be bumped again

richfitz (changes requested):

R/git.R
r-ash: It absolutely is

NEWS.md
r-ash: Thanks, fixed

R/remote.R
r-ash: Yes, that function does it already for us too. I've simplified this PR a bit

richfitz (approved):
nice, this is super tidy

PR #267, opened 2021-02-10
Don't save output with colour and strip colour codes from testing

r-ash:
This PR will
* Not save `output.log` with colour encoding
* Strip it elsewhere where it was causing test failures

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -217,7 +217,7 @@ capture_log <- function(expr, filename) {
     close(con)
   })
   handle_message <- function(e) {
-    cat(e$message, file = stdout())
+    cat(crayon::strip_style(e$message), file = stdout())
   }
   handle_error <- function(e) {
     calls <- sys.calls()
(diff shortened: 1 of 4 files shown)


R/util.R
```diff
@@ -217,15 +217,16 @@ capture_log <- function(expr, filename) {
...
+    cat(crayon::strip_style(e$message), file = stdout())
   }
   handle_error <- function(e) {
     calls <- sys.calls()
     str <- utils::limitedLabels(calls[seq_len(length(calls) - 2L)])
     lab <- format(seq_along(str))
     str <- gsub("\n", sprintf("\n%s", strrep(" ", nchar(lab)[[1]] + 2)), str)
     trace <- paste(sprintf("%s: %s\n", lab, str), collapse = "")
-    cat(sprintf("Error: %s\nTraceback:\n%s", e$message, trace),
+    cat(crayon::strip_style(sprintf("Error: %s\nTraceback:\n%s",
```
richfitz, line 228: I don't think there should be style by this point (though this is likely safe)

richfitz (changes requested):
Can you please bump the version number so that the release build doesn't fail?

R/util.R
r-ash: Thanks, have removed that

richfitz (approved):

Issue #259, opened 2020-12-24
error specifying dependency id with use_draft = "newer"


jeffeaton:
When I specify a dependency on a specific task id, and also specify `orderly_run(..., use_draft = "newer")`, I get an error message:
```r
> id2b <- orderly::orderly_run("task2", root = temproot, use_draft = "newer")
[ name       ]  task2
Error in if (draft) path_draft else path_archive : 
  argument is not interpretable as logical
```

Here is the example `orderly.yml`:
```
script: script.R

artefacts:
  - data:
      description: Another file
      filenames:
        - something.txt

depends:
 - task1:
    id: latest
    use:
      hello-new.txt: hello.txt
 - task1:
    id: 20201224-155142-2a0c2fde
    use:
      hello-orig.txt: hello.txt
```

It works okay if instead I specify `draft: true` in the `orderly.yml` and run without the `use_draft=` argument, but I get a warning that `draft: true` is deprecated:

```
...
depends:
 - task1:
    id: latest
    draft: true
    use:
      hello-new.txt: hello.txt
 - task1:
    id: 20201224-155142-2a0c2fde
    use:
      hello-orig.txt: hello.txt
```

```r
> orderly::orderly_run("task3", root = temproot)
Warning: Using 'draft:' within an ordery.yml is deprecated and will be removed
in a future version of orderly.  Please use the 'use_draft' argument to
control draft usage. If you want to use a recent version of a report
that you are developing simultaneously, use_draft = "newer" will
probably do what you want.
[ name       ]  task3
[ id         ]  20201224-162108-c7c535b3
[ depends    ]  task1@20201224-161330-53fe599a:hello.txt -> hello-new.txt
[ ...        ]  task1@20201224-161326-d0326f1b:hello.txt -> hello-orig.txt
[ start      ]  2020-12-24 16:21:08
```
-----
Here's a full reproducible example:
```r
temproot <- tempfile()

orderly::orderly_init(temproot)

## Create a task1

orderly::orderly_new("task1", root = temproot)

writeLines("writeLines(\"hello world\n\", \"hello.txt\")",
           file.path(temproot, "src/task1/script.R"))

writeLines(c("script: script.R",
             "",
             "artefacts:",
             "  - data:",
             "      description: A file",
             "      filenames:",
             "        - hello.txt"),
           file.path(temproot, "src/task1/orderly.yml"))

id1 <- orderly::orderly_run("task1", root = temproot)
orderly::orderly_commit(id1, root = temproot)


[...]

PR #269, opened 2021-02-10
VIMC-4579: Enable orderly_latest to find latest from draft or archive

r-ash:
Addresses #259 

The issue in #259 was because resolving dependencies can take e.g. `draft = "newer"` which works with finding the report in `orderly_find_report` and searching in `orderly_search` but it doesn't work when checking if the report we found is the latest report using `orderly_latest`. To fix I have added support to `orderly_latest` to take draft `newer`, `always` or `never`.

### DIFF
--- a/R/query.R
+++ b/R/query.R
@@ -143,36 +143,11 @@ orderly_latest <- function(name = NULL, root = NULL, locate = TRUE,
                            draft = FALSE, must_work = TRUE) {
   config <- orderly_config(root, locate)
 
-  get_latest_archive <- function() {
-    if (is.null(name)) {
-      d <- orderly_list2(FALSE, config, FALSE)
-      ids <- d$id
-      path <- file.path(path_archive(config$root), d$name)
-    } else {
-      path <-
-        file.path(path_archive(config$root), name)
-      ids <- orderly_list_dir(path, check_run_rds = FALSE)
-    }
-    ids
-  }
-  get_latest_draft <- function() {
-    if (is.null(name)) {
-      d <- orderly_list2(TRUE, config, FALSE)
-      ids <- d$id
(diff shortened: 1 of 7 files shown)


R/query.R
```diff
@@ -139,22 +143,25 @@ orderly_latest <- function(name = NULL, root = NULL, locate = TRUE,
...
-    ids <- d$id
-    path <-
-      file.path((if (draft) path_draft else path_archive)(config$root), d$name)
-  } else {
-    path <-
-      file.path((if (draft) path_draft else path_archive)(config$root), name)
-    ids <- orderly_list_dir(path, check_run_rds = draft)
+  draft <- query_check_draft(draft)
+  path_funcs <- switch(draft,
+                       always = c(list_draft),
```
richfitz, line 148: I think that `c()` here should be `list()`

R/query.R
```diff
@@ -139,22 +143,25 @@ orderly_latest <- function(name = NULL, root = NULL, locate = TRUE,
...
+                       never = c(list_archive),
+                       newer = c(list_draft, list_archive))
+  what <- switch(draft,
+                 always = "draft",
+                 never = "archive",
+                 newer = "draft or archive")
+
+  ids <- c()
```
richfitz, line 156: `c()` is `NULL` but I think here you mean `character(0)`

richfitz: alternatively can you do:

```
ids <- unlist(lapply(path_funcs, function(f) f(name, config)))
```

richfitz (changes requested):

R/query.R
r-ash: > `c()` is `NULL` but I think here you mean `character(0)`

Oops yep, thanks - I think I must have done this many times before

richfitz (approved):

Issue #265, opened 2021-01-21
orderly_cleanup() does not delete draft if trailing / in name


jeffeaton:
If I specify a task by name with a trailing `/`, `orderly_cleanup("task1/")` does not remove the task and does not return any message.

Two suggestions for behaviour:
* It would be nice for `orderly_cleanup()` to sanitise the name so `orderly_cleanup("task1/")` does the same as `orderly_cleanup("task1").
* It would be nice to return a warning or an error message if the task is not found, e.g. `orderly::orderly_cleanup("taskDOESNTEXIST")` gives me a warning or error.

Here is a small example:

```r
temproot <- tempfile()

orderly::orderly_init(temproot)

## Create a task1

orderly::orderly_new("task1", root = temproot)

writeLines("writeLines(\"hello world\n\", \"hello.txt\")",
           file.path(temproot, "src/task1/script.R"))

writeLines(c("script: script.R",
             "",
             "artefacts:",
             "  - data:",
             "      description: A file",
             "      filenames:",
             "        - hello.txt"),
           file.path(temproot, "src/task1/orderly.yml"))

id1 <- orderly::orderly_run("task1", root = temproot)

orderly::orderly_list_drafts(root = temproot)

## If task name has a trailing /, orderly_cleanup() does nothing and returns no message.
orderly::orderly_cleanup("task1/", root = temproot)
orderly::orderly_list_drafts(root = temproot)

## This removes the task
orderly::orderly_cleanup("task1", root = temproot)
orderly::orderly_list_drafts(root = temproot)
```

-----
Here's the output I see:

```r
> temproot <- tempfile()
> orderly::orderly_init(temproot)
Now, edit the file 'orderly_config.yml' within '/var/folders/fy/vmz3dx2x5bg1f8vy3z0yh5fr0000gn/T//RtmpIVbMak/file15d113a3c8ce'
[1] "/var/folders/fy/vmz3dx2x5bg1f8vy3z0yh5fr0000gn/T//RtmpIVbMak/file15d113a3c8ce"
> orderly::orderly_new("task1", root = temproot)
Created report at '/private/var/folders/fy/vmz3dx2x5bg1f8vy3z0yh5fr0000gn/T/RtmpIVbMak/file15d113a3c8ce/src/task1'
Edit the file 'orderly.yml' within this directory
> writeLines("writeLines(\"hello world\n\", \"hello.txt\")",
+            file.path(temproot, "src/task1/script.R"))
> writeLines(c("script: script.R",
[...]

PR #270, opened 2021-02-10
VIMC-4580: Make orderly_cleanup chattier and sanitise report name

r-ash:
* This will now print out how many draft reports, csv files and rds files are going to be deleted
* And sanitise report name before running cleanup (i.e. strip / and strip preceeding `src`)

I think this should address #265

### DIFF
--- a/R/cleanup.R
+++ b/R/cleanup.R
@@ -60,6 +60,12 @@ orderly_cleanup_drafts <- function(config, name = NULL, failed_only = FALSE) {
   if (failed_only) {
     p <- p[!file.exists(path_orderly_run_rds(p))]
   }
+  msg <- sprintf("Found %s draft %s", length(p),
+                 ngettext(length(p), "report", "reports"))
+  if (!is.null(name)) {
+    msg <- paste(msg, sprintf("for report name '%s'", name))
+  }
+  orderly_log("clean", msg)
   if (length(p) > 0L) {
     orderly_log(if (failed_only) "prune" else "clean", p)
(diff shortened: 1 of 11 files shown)


R/util.R
```diff
@@ -925,3 +925,13 @@ lock_bindings <- function(syms, env) {
   }
   invisible(NULL)
 }
+
+clean_report_name <- function(name) {
```
richfitz, line 929: This duplicates logic in the increasingly misnamed `orderly:::orderly_develop_location`; can we use that instead?

R/cleanup.R
```diff
@@ -85,12 +92,18 @@ orderly_cleanup_data <- function(config) {
 
   csv <- orderly_db("csv", config, FALSE)
   drop_csv <- setdiff(csv$list(), used)
+  msg <- sprintf("Found %s csv %s", length(drop_csv),
```
richfitz, line 95: oh man totally forgot about all this old stuff. We should tidy this up at some point...

richfitz (changes requested):

R/util.R
r-ash: This PR pulls that file name cleaning logic out of `orderly_develop_location` into this new function. Then `orderly_develop_location` calls `clean_report_name`. Or you think `orderly_ cleanup` should call `orderly_develop_location`?

richfitz: Sorry, my mistake - you're right of course

richfitz (approved):

PR #278, opened 2021-04-13
MRC-2321: max-depth arg to orderly_graph now limits depth of tree returned

r-ash:
instead of throwing an error

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # orderly 1.2.32
 
-* `orderly_graph` arg `max_depth` now truncates depth of the tree instead of throwing error and arg `recursion_depth` added which works how `max_depth` used to
+* `orderly_graph` arg `max_depth` now truncates depth of the tree instead of throwing error and arg `recursion_limit` added which works how `max_depth` used to
 
 # orderly 1.2.30
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,8 @@
-# orderly 1.2.39
+# orderly 1.2.32
+
+* `orderly_graph` arg `max_depth` now truncates depth of the tree instead of throwing error and arg `recursion_depth` added which works how `max_depth` used to
+
+# orderly 1.2.30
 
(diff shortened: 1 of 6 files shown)


tests/testthat/test-dependency.R
```diff
@@ -444,3 +444,160 @@ test_that("detect loop", {
...
+  expect_length(child_1$children, 0)
+
+  expect_error(orderly_graph("depend3", root = path, use = "src",
+                             propagate = FALSE, show_all = TRUE,
+                             direction = "upstream", max_depth = 1,
+                             recursion_limit = 0),
+               "The tree is very large or degenerate")
+})
+
+test_that("dependency latest with search query", {
```
r-ash, line 580: Added this test too to convince myself that this is how it is working, figured it would be generally useful as I don't think this case is covered

NEWS.md
```diff
@@ -1,4 +1,8 @@
-# orderly 1.2.39
+# orderly 1.2.32
+
+* `orderly_graph` arg `max_depth` now truncates depth of the tree instead of throwing error and arg `recursion_depth` added which works how `max_depth` used to
```
richfitz, line 3: this should be `recursion_limit`

R/dependencies.R
```diff
@@ -91,23 +94,24 @@
 ##'                                  direction = "upstream")
 orderly_graph <- function(name, id = "latest", root = NULL, locate = TRUE,
                           direction = "downstream", propagate = TRUE,
-                          max_depth = 100, show_all = FALSE,
-                          use = "archive") {
+                          max_depth = 100, recursion_limit = 100,
```
richfitz, line 97: default for max_depth should be Inf I think

richfitz (changes requested):
thanks - looks good aside from two extremely minor things

NEWS.md
r-ash: oops, thanks!

richfitz (approved):

PR #279, opened 2021-04-15
Add report_version_instance table to store info about db instances report ran against

r-ash:
Does this need any migration test? I think the specific migration tests are for migrations to the archive? But here this is just db update which will be managed automatically by running `orderly_rebuild` I believe?

### DIFF
--- a/R/db2.R
+++ b/R/db2.R
@@ -444,13 +444,24 @@ report_data_import <- function(con, name, id, config) {
 
   ## DB instance:
   if (!is.null(dat_rds$meta$instance)) {
-    report_version_instance <- data_frame(
-      report_version = id,
-      type = names(dat_rds$meta$instance),
-      instance = unname(unlist(dat_rds$meta$instance))
-    )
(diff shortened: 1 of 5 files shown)


R/db2.R
```diff
@@ -442,6 +442,29 @@ report_data_import <- function(con, name, id, config) {
                       append = TRUE)
   }
 
+  ## DB instance:
+  if (!is.null(dat_rds$meta$instance)) {
+    ## Only add row where instance is set
+    build_row <- function(type) {
```
richfitz, line 448: I think there's a little more coersion here than is safe, and in particular I think this might not work as expected on 3.6.x due to treatment of factors. Consider:

```
if (is.null(instance)) {
```

richfitz: (with data_frame being our internal wrapper that does `data.frame(..., stringsAsFactors = FALSE)`

richfitz (changes requested):

R/db2.R
r-ash: That's way nicer, thanks

richfitz (approved):

PR #284, opened 2021-05-04
mrc-2337: Remove support for running workflow

r-ash:
This PR will
* Remove `orderly_workflow` function and remove workflow runner from CLI
* Expose `workflow_id` arg on cli runner, expect that `orderly.server` will start passing this in. It just gets stored in database and as metadata in the rds.

I think I probably still need to do a migration as this drops the `name` column from the orderly db


R/db2.R
```diff
@@ -429,16 +429,13 @@ report_data_import <- function(con, name, id, config) {
...
-        name = dat_rds$meta$workflow$name
-      )
+    if (nrow(DBI::dbGetQuery(con, sql_batch, dat_rds$meta$workflow)) == 0L) {
+      workflow <- data_frame(id = dat_rds$meta$workflow)
       DBI::dbWriteTable(con, "workflow", workflow, append = TRUE)
     }
     report_version_workflow <- data_frame(
       report_version = id,
-      workflow_id = dat_rds$meta$workflow$id
+      workflow_id = dat_rds$meta$workflow
```
richfitz, line 438: Does this mean we need a metadata migration?

richfitz (changes requested):

R/db2.R
r-ash: Discussed this, no one has used it so we can get away with not writing a migration

richfitz (approved):

PR #289, opened 2021-06-09
Prepare for CRAN update

richfitz:
* Fix URLs (some redirects, other were at repos or links that were private)
* Make sure we consistently skip on CRAN where needed
* Tidy up some dockerfile issues
* Fix spelling
* Fix test that wrote to the test dir (297ea43)


tests/testthat/test-info.R
```diff
@@ -28,7 +28,8 @@ test_that("can retrieve parameter info", {
 })
 
 test_that("can retrieve info from failed run", {
-  path <- prepare_orderly_git_example()
+  skip_if_no_git()
```
r-ash, line 31: Can get rid of this line

r-ash (changes requested):
Looks good, one minor thing I noticed. Otherwise good to go

r-ash (approved):

PR #290, opened 2021-06-24
Add pkgdown URL to DESCRIPTION

maelle:
Also, there's no pkgdown config file yet but it'd be good to have one and to add the URL to it as well (for e.g. Twitter metadata). :slightly_smiling_face:

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -20,7 +20,7 @@ Authors@R: c(person("Rich", "FitzJohn", role = c("aut", "cre"),
              person("James", "Thompson", role = "aut"),
              person("Imperial College of Science, Technology and Medicine",
                     role = "cph"))
-URL: https://github.com/vimc/orderly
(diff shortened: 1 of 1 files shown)


richfitz:
Thanks @maelle - I missed this. I'll merge this into another active branch we're working on given it's just a metadata change.

Could you please point me to an example pkgdown config? It's not clear what we're missing by not having one, but that is mostly my ignorance...

maelle:
A config file is good for many things like grouping functions on the reference index but for the URL itself in the configuration see https://pkgdown.r-lib.org/dev/reference/build_site.html#yaml-config

With pkgdown dev version you can share a config file in a template package which might be useful for e.g. author configuration. https://pkgdown.r-lib.org/dev/articles/customization.html#sharing-pkgdown-styles-template-packages-1

richfitz (approved):

PR #292, opened 2021-07-07
Allow partial pulling of orderly trees

richfitz:
This PR extends the remote pull functions (orderly_pull_archive and orderly_pull_remote) to allow grabbing just the leaf report, creating a metadata store that can be used to build the db.

Not implemented, but could be, or pushed to future tickets:

* Optional depth to recurse to. This is easy to implement but I wonder if it's needed
* Functions to resolve partial archives, replacing metadata with real outputs. Not totally sure this is needed either
* Functions to remove redundant metadata


r-ash (approved):
This looks good. We might need a function to resolve partial archive if say a user has report A and B which depends on A. They want to write a report using report B as a dependency so they pull B, which pulls metadata for A. They then want to subsequently write some other report which depends on report A - they will need to resolve the partial archive to do that?

richfitz:
> This looks good. We might need a function to resolve partial archive if say a user has report A and B which depends on A. They want to write a report using report B as a dependency so they pull B, which pulls metadata for A. They then want to subsequently write some other report which depends on report A - they will need to resolve the partial archive to do that?

They will; at the moment it'd look like this:

```
dat <- prepare_orderly_remote_example()
id3 <- orderly_run("depend", root = dat$path_remote, echo = FALSE)
orderly_commit(id3, root = dat$path_remote)
orderly_pull_archive("depend", root = dat$config, remote = dat$remote,
                     recursive = FALSE)

orderly_run("depend", root = dat$config)
## [ name       ]  depend
## Error in orderly_find_report(id, name, config, draft = use_draft, must_work = TRUE) : 
##   Did not find archive report example:latest

orderly_pull_dependencies("depend", root = dat$config, remote = dat$remote)
## [ depends    ]  depend has 1 dependency
## [ pull       ]  example:20210707-134836-9dffa6d8
## [ import     ]  example:20210707-134836-9dffa6d8

orderly_run("depend", root = dat$config)
## [ name       ]  depend
## (snip)
## [ end        ]  2021-07-07 13:50:00
## [ elapsed    ]  Ran report in 0.01128268 secs
## [ artefact   ]  mygraph.png: 589bd1d5b44a486dfc53c9e4bb0d4fde
## [ ...        ]  output.rds: 6733ee6352f32efc7834e1d9f4538bc8
## [1] "20210707-135000-96d471fe"
```

Did you have something else in mind?

PR #298, opened 2021-08-19
Workaround for zip_list on large files

richfitz:
This is temporary, and we may not merge it depending on how the fix to zip goes (see https://github.com/r-lib/zip/pull/79)

Let's merge for now and remove once zip is updated on CRAN


weshinsley (approved):
Looks great.
We wait for https://github.com/r-lib/zip/pull/79 ...!

Issue #301, opened 2021-09-17
Rscript.exe crashes if orderly_bundle_run path is not found


weshinsley:
It appears that the below on Windows crashes RScript.exe (4.1.0) with a null pointer exception, even within a tryCatch:-
`tryCatch(zip::zip_list("doesnt_exist.zip"), error = function(e) message("ERROR"))` 

So a call to orderly_bundle_run with a bad path, that calls orderly_bundle_info, then reaches bundle.R:217

```
zip_list2 <- function(path) {
  tryCatch(zip::zip_list(path),
           error = function(e)
             tryCatch(zip_list_base(path),
                      error = function(e2) stop(e)))
}
```

Probably this is in the air re the potential fix to zip_list. But we could consider some earlier validation on the path.

weshinsley:
Hmm. I can get the above error on a cluster node, but not reproduce it locally...

weshinsley:
Fixed in the above PR - nb this was not a bug in orderly, but in the zip package, which I think is fixed with https://github.com/r-lib/zip/pull/79

It remains unclear why I cannot provoke the crash on my Win 10 desktop, but it is consistent on a Win 2012 R2 server node, but the crash happens clearly in the call to the C code within the zip library.

PR #302, opened 2021-09-20
Fix reported CRAN check failure

richfitz:
This PR fixes the current failing builds here: https://cran.r-project.org/web/checks/check_results_orderly.html

My belief as to what is going on is in the comment in test-run.R: knitr now requires that anyone knitting a markdown (rather than sweave) document includes the dependency on `markdown` even though that is clearly a knitr dependency (see https://github.com/yihui/knitr/issues/1864), then there's a workaround for this unfortunate decision in knitr which seems to break our package (https://github.com/yihui/knitr/commit/9a80a00). However, this breaks *differently* on CRAN machines and ours because this workaround tries not to error and instead silently writes an empty file, so the test fails.

Having updated things to include knitr's dependency among ours (as I guess hundreds of packages have had to?) I've also just removed this test from within those running on CRAN as I can't be sure that this is real reason - it's still unexplained why only devel and patched builds are failing but that might be lag. In any case given we're in workaround-ception territory here, it feels best not to expose the package to removal based on this situation.

Also includes a fix for #301, by including an explicit check for the path. This should give a better error for the user even if the zip package did throw its error.

Also fixes vimc-4345

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -98,10 +98,6 @@ jobs:
         run: |
           ./scripts/copy_reference
 
-      - name: Setup tmate session
-        if: runner.os == 'Windows'
-        uses: mxschmitt/action-tmate@v3
-
       - name: Check
         env:
           _R_CHECK_CRAN_INCOMING_REMOTE_: false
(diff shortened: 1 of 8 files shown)


weshinsley (approved):
Looks good - frustrating to have to do this because of external forces...
Tested on Windows - all good
Also verified that on the server node, this fixes issue #301 as far as orderly usage goes.

PR #306, opened 2021-11-08
Update orderly demo script to include dependency reports in git example

r-ash:
This simplifies the orderly test setup for OrderlyWeb
The `demo` dir will contain an orderly root with git and with reports have use dependencies `use_dependency` and `use_dependency_2`

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: orderly
 Title: Lightweight Reproducible Reporting
-Version: 1.4.3
+Version: 1.4.4
 Description: Order, create and store reports from R.  By defining a
     lightweight interface around the inputs and outputs of an
     analysis, a lot of the repetitive work for reproducible research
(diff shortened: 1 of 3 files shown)


r-ash:
I've made a further change to this, now what we have is a single repo with 2 branches master and other
* master - exactly the same as previous `git` master
* other - same as previous `git` master but with all additional demo reports (including ones with dependencies) can see these here https://github.com/vimc/orderly/tree/master/inst/examples/demo/src

This is going to make orderly.server tests fail so if this look good to you, let me know and I will update orderly.server

r-ash:
Ok further change! We need to think more about what we actually need from the testing so this should be the minimal change which gives you some reports with dependencies in the `git` demo repo. This will
* Add all additional demo reports to the `other` branch of the `git` demo repo, including `use_dependency` and `use_dependency_2`

EmmaLRussell (approved):
Have tested this out in an OW branch, and it does not cause any build issues (last build failed for unrelated reason), and new reports are visible in 'other' branch

PR #308, opened 2021-12-22
Return more informative error from deduplication

r-ash:
I've put this error in as soon as we list the paths to fail early. And not removed the 
```
stopifnot(all(vlapply(split(files, files$hash), function(x)
    all(x$inode_first == x$inode[[1]]))))
```
which is what Oli has been seeing error previously. I can remove it if you think no longer relevant

### DIFF
--- a/R/deduplicate.R
+++ b/R/deduplicate.R
@@ -83,6 +83,7 @@ orderly_deduplicate_info <- function(config) {
                           files$filename,
                           fsep = "/")
 
+  paths <- file.path(config$root, "archive", files$path)
   ## Information about the physical files, so we can work out which
   ## files are already hardlinked
   files <- cbind(
--- a/R/deduplicate.R
+++ b/R/deduplicate.R
@@ -79,6 +79,13 @@ orderly_deduplicate_info <- function(config) {
                           fsep = "/")
 
   paths <- file.path(config$root, "archive", files$path)
+  exists <- vlapply(paths, file.exists)
+  if (any(!exists)) {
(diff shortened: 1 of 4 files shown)


R/deduplicate.R
```diff
@@ -79,6 +79,13 @@ orderly_deduplicate_info <- function(config) {
                           fsep = "/")
 
   paths <- file.path(config$root, "archive", files$path)
+  exists <- vlapply(paths, file.exists)
```
richfitz, line 82: I wonder about looking for the existence of `file.path(config$root, "metadata")` (or perhaps that there is at least one file within that directory); that would be stricter and easier to check?  What do you think?

R/deduplicate.R
```diff
@@ -79,6 +79,13 @@ orderly_deduplicate_info <- function(config) {
...
+  if (any(!exists)) {
+    stop(paste0("Cannot deduplicate archive as database references files ",
```
richfitz, line 84: This is nicer to construct with `paste` and drop the trailing whitespace at the end of each string

richfitz (changes requested):

R/deduplicate.R
r-ash: If we want to fail with a specific message for people who have done a pull with `recursive = FALSE` then this works. People could still be in a state where they get a not very informative error later though from the stopifnot call. Say if they have deleted a report but not rebuilt the database.

richfitz (approved):

PR #310, opened 2022-01-17
Add an example report to orderly demo which includes a default parameter

r-ash:
Requested by @LekanAnni for testing something in orderly web. Atm there is no demo report with a default parameter.

I've checked this doesn't affect any of the orderly.server tests


LekanAnni (approved):
Thanks

richfitz (approved):

PR #314, opened 2022-04-22
mrc-3167: Add function to cancel reports running on remote

r-ash:
This PR will
* Add function `orderly_cancel_remote` which will call `kill` from orderly.server to cancel/kill a report on the remote with its key.

Couple of questions
* Should exporting `kill` function become a check in `implements_remote`? I need to add to sharepoint remote still to explicitly throw an error if trying to call `kill` there
* Do we want to do any parsing of the return? This pretty much just returns the output from orderly.server straight to users
* What do you think of the name? I tried `orderly_cancel_remote` as close to `orderly_run_remote`
* No example for most of the remote functions, is that just the way it is because we don't have a real remote setup for the tests. I could add an example which doesn't get run

https://github.com/vimc/orderlyweb/pull/18 needs to be merged before this will work


NAMESPACE
```diff
@@ -9,6 +9,7 @@ export(orderly_bundle_list)
 export(orderly_bundle_pack)
 export(orderly_bundle_pack_remote)
 export(orderly_bundle_run)
+export(orderly_cancel_remote)
```
richfitz, line 12: should this be `orderly_remote_cancel` to mirror `orderly_remote_status` or stay `orderly_cancel_remote` to mirror `orderly_run_remote`? Naming things is hard

R/remote.R
```diff
@@ -662,3 +662,29 @@ pull_info <- function(name, id, root, locate, remote, parameters) {
...
+##' @return List with names as report keys and values are lists containing
+##'   * `killed` - boolean TRUE if report successfully cancelled, FALSE
+##'   otherwise
+##'   * `message` - string detailing reason why cancellation failed
+##'
+##' @export
+orderly_cancel_remote <- function(keys, root = NULL, locate = TRUE,
+                                  remote = NULL) {
+  remote <- get_remote(remote, orderly_config(root, locate))
+  out <- lapply(keys, remote$kill)
```
richfitz, line 687: If you don't want to make this part of the "implements remote" thing, fall back here:

```
if (is.null(remote$kill)) {
  stop(sprintf("your remote (there's a way of getting the type I believe) does not implement kill")
} 
```

that's probably the tidy thing to do as we don't want to implement this for the path remote

richfitz, line 687: If you don't want to make this part of the "implements remote" thing, fall back here:

```
if (is.null(remote$kill)) {
  stop(sprintf("your remote (there's a way of getting the type I believe) does not implement kill")
} 
```

that's probably the tidy thing to do as we don't want to implement this for the path remote

richfitz (changes requested):
couple of small comments

NAMESPACE
r-ash: We have more functions which are `orderly_x_remote`, though they do tend to be when there is a local equivalent - which there isn't here. I think `orderly_cancel_remote` feels more natural wording to me so I am happy to stick with that, I don't mind switching it over if you prefer though

richfitz (approved):

PR #315, opened 2022-04-26
Remove demo test dir and run demo reports on other branch

r-ash:
**N.B. We should not merge this until orderly-web tests are updated and working again, it is going to potentially break lots of stuff**

This PR will
* Remove `demo` dir from `create_orderly_demo.sh` test setup which is used by orderly-web
* Run the demo reports on the `other` branch in the `git` directory (so the report_version required have records in the database)

This does change some of the git state in `git` directory. Namely previously we had

Locally
master: initial-import
other: initial-import  -- add-other

In upstream
master: initial-import -- orderly
other: initial-import -- add-other

We now have

Locally
master:   initial-import
other:  initial-import  -- add-other -- run demo reports

In upstream
master: initial-import -- orderly
other: initial-import -- add-other -- run demo reports

### DIFF
--- a/R/testing.R
+++ b/R/testing.R
@@ -247,8 +247,8 @@ unzip_git_demo <- function(path = tempfile()) {
   path
 }
 
-prepare_orderly_git_example <- function(path = tempfile(), run_report = FALSE,
-                                        branch = "master") {
+prepare_orderly_git_example <- function(path = tempfile(), run_minimal = FALSE,
+                                        run_other = FALSE, branch = "master") {
   path_upstream <- file.path(path, "upstream")
   unzip_git_demo(path)
   unzip_git_demo(path_upstream)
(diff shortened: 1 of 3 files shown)


LekanAnni:
Can we also have some reports published, as well as data in changelog and Depends tables linking to any of these reports?

r-ash:
> Can we also have some reports published, as well as data in changelog and Depends tables linking to any of these reports?

Yeah, so this is running the reports in the same way that demo was so those tables should all be populated.

EmmaLRussell (commented):
> Locally
master: initial-import
other: initial-import -- add-other -- run demo reports

> In upstream
master: initial-import -- orderly
other: initial-import -- add-other -- run demo reports

This is referring to the git commits? So we'll now have two commits on master branch and three on other branch? I think that's fine but it will mean we'll need to update some of our tests, as I think e.g. Selenium test checks for number of commits available in the drop-down. 

Generally, this approach sounds like it should work well to me, but will see how fixing it up in OW goes!

r-ash:
> Locally
> master: initial-import
> other: initial-import -- add-other -- run demo reports

Yes this is
<branch name>: <commit msg 1> -- <commit msg 2> --etc.

PR #317, opened 2022-05-06
Fix documentation for where to store templates

BlackEdder:
There seemed to be a small error in the documentation. It said templates should go in the "templates" directory, but `has_template()` actually looks in the __template__ directory.


richfitz (approved):
Thanks! Good spot :)

PR #318, opened 2022-05-24
Allow control over default branch

richfitz:
This will be used by orderly.server to control a few hard-coded "master" places


r-ash (approved):
Looks good to me! Will leave open until we can see working on some deployed version

PR #319, opened 2022-05-27
vimc-6487: Tidy unpredictable search behaviour with orderly_pull_archive and orderly_push_archive

r-ash:
This PR does 2 things
* Supports search query as `id` arg to `orderly_push_archive` (which is what the docs suggested was already working)
* Removes `parameter` arg from `orderly_pull_archive`, saw users trying to use this as `orderly_pull_archive("report_name", parameters = list(nmin = 0.5))` to pull the latest report where a param had a value. I think the `parameters` arg to this is generally quite confusing and I'm not sure what the use case is here? In `orderly_pull_dependencies` it makes sense as the search query is set in the `orderly.yml` itself whereas with `orderly_pull_archive` the user sets the query term so they could just write `latest(parameter:nmin == 0.5)` instead of using `id = "latest(parameter:nmin == n)", parameters = list(nmin = 0.5)`. Could be a use case for that that I have not thought about but I think fairly confusing taking both args atm.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: orderly
 Title: Lightweight Reproducible Reporting
-Version: 1.4.8
+Version: 1.4.9
 Description: Order, create and store reports from R.  By defining a
     lightweight interface around the inputs and outputs of an
     analysis, a lot of the repetitive work for reproducible research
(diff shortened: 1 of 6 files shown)


richfitz:
Sorry, will look at this shortly. Just leaving some notes here for myself - this seems related to

* https://github.com/vimc/orderly/pull/262 / https://mrc-ide.myjetbrains.com/youtrack/issue/VIMC-4492
* https://mrc-ide.myjetbrains.com/youtrack/issue/VIMC-4506

PR #320, opened 2022-10-06
mrc-3665 way for orderly web to control demo data

hillalex:
This adds a new entrypoint that exposes the function for generating an orderly demo based on a source directory and a `demo.yml` file. See this in use in OrderlyWeb: https://github.com/vimc/orderly-web/pull/479

`R/testing.R` is a bit of a tangle tbh, but rather than unpicking it I've just added a new function for what we want in OW.


inst/run_orderly_demo.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
```
hillalex, line 1: I think the existing `create_orderly_demo` script is still in use by things other than OW (e.g. for testing in this repo), so I've left that as is and just added a new one for the usage we want in OW.

r-ash: Looks like `create_orderly_demo` script is used in test set in many places, like montagu-proxy, montagu-reporting-api, montagu-webapps. If we are not using `create_orderly_demo` in orderly-web any more then possibly worth backing out commit https://github.com/vimc/orderly/pull/320/commits/d7922bd892ee4cbe1be1e9169d1de89b823c05a1 from this PR to avoid potentially breaking the other components which use it

r-ash (commented):
New function for running git example from source looks good. Small comment about the changes to create_orderly_demo script.

It would be nice to review all the setup functions in `R/testing.R` at some point as they have all expanded since first being added and I think probably have some overlap in functionality. The differences between them are quite confusing. Can see that is out of scope for what you want to do in this ticket though

inst/run_orderly_demo.sh
hillalex: Ok yep, let's do that.

hillalex:
> New function for running git example from source looks good. Small comment about the changes to create_orderly_demo script.
> 
> It would be nice to review all the setup functions in `R/testing.R` at some point as they have all expanded since first being added and I think probably have some overlap in functionality. The differences between them are quite confusing. Can see that is out of scope for what you want to do in this ticket though

Agree that adding a further function feels like adding more complication on top of an already confusing collection of code. I'd be happy to have a go at refactoring the whole file in another ticket.

hillalex:
I'm a little puzzled as to why the windows R-CMD-check is failing, given that it seems to be passing for other branches...

richfitz:
quite likely it's failing because openssl has been updated on cran, GHA is trying and failing to install it from source - probably something in that vignette is using openssl indirectly

richfitz:
indeed that is what's going on:
```
Error: Error: processing vignette 'orderly.Rmd' failed with diagnostics:
there is no package called 'openssl'
--- failed re-building 'orderly.Rmd'
--- re-building 'patterns.Rmd' using rmarkdown
--- finished re-building 'patterns.Rmd'
--- re-building 'remote.Rmd' using rmarkdown
--- finished re-building 'remote.Rmd'
SUMMARY: processing the following files failed:
  'bundles.Rmd' 'orderly.Rmd'
Error: Error: Vignette re-building failed.
```

R/testing.R
```diff
@@ -51,6 +51,16 @@ create_orderly_demo <- function(path = tempfile(), quiet = FALSE,
 }
 
 
+prepare_git_example_from_source <- function(source_path, path) {
```
richfitz, line 54: Can you add a test for this within the testthat tests please?

inst/run_orderly_demo.sh
richfitz, line 1: Generally I think it's nicer to avoid using extensions on scripts as then we can reimplement them in another language later without having to update uses. Your editor should pick up syntax highlighting properly based on the first line

richfitz (changes requested):
Don't worry about the windows failing build, this is affecting lots of things atm and it's very unlikely this set of changes will break that build specifically

r-ash:
> quite likely it's failing because openssl has been updated on cran, GHA is trying and failing to install it from source - probably something in that vignette is using openssl indirectly

There is a fix for this in #321 https://github.com/vimc/orderly/pull/321/files#diff-9c940e8ad2b7bc4c26ec3da57b94bc00e73e2166cfed689da51a4c59bcc0a310R80

docker/Dockerfile
```diff
@@ -53,6 +53,7 @@ COPY . /orderly
 RUN R CMD INSTALL /orderly && \
   Rscript -e 'orderly:::write_script("/usr/bin")' && \
   cp /orderly/inst/create_orderly_demo.sh /usr/bin/ && \
+  cp /orderly/inst/run_orderly_demo /usr/bin/ && \
```
hillalex, line 56: I've left the old script `create_orderly_demo.sh` with the extension, since it could be in use by other repos

hillalex:
> 

Huh, weird. Also looks like there are some strange test failures, will investigate.

richfitz (approved):

PR #321, opened 2022-10-19
Trigger orderly.server downstream build after check completed

r-ash:


.github/workflows/R-CMD-check.yaml
```diff
@@ -110,3 +116,19 @@ jobs:
...
+
+    needs: R-CMD-check
+
+    steps:
+      - uses: convictional/trigger-workflow-and-wait@v1.6.1
+        with:
+          owner: vimc
+          repo: orderly.server
+          ref: master
+          github_token: ${{ secrets.REPO_DISPATCH_TOKEN }}
```
r-ash, line 132: Sadly we need the generated token here

.github/workflows/R-CMD-check.yaml
```diff
@@ -110,3 +116,19 @@ jobs:
...
           path: check
+
+  check-orderly-server:
+
+    runs-on: ubuntu-20.04
+
+    needs: R-CMD-check
+
+    steps:
+      - uses: convictional/trigger-workflow-and-wait@v1.6.1
```
r-ash, line 127: Managed to find an existing step which will trigger it and wait. See https://github.com/marketplace/actions/trigger-workflow-and-wait

buildkite/pipeline.yml
```diff
@@ -33,3 +33,10 @@ steps:
     branches: "master"
     build:
       branch: master
+
+  - trigger: "orderly-dot-server-test"
+    label: ":rocket: orderly.server (from orderly) :cloud:"
+    build:
+      branch: !master
```
r-ash, line 40: quoting this causes it to interpret `!master` as the branch name instead of as not `master`

hillalex (commented):
Does the current set up cause an `orderly.server` docker image, based on this branch, to be erroneously tagged as "master" and pushed to docker hub?

r-ash:
> Does the current set up cause an `orderly.server` docker image, based on this branch, to be erroneously tagged as "master" and pushed to docker hub?

It shouldn't do, I added a new buildkite pipeline to orderly.server which doesn't push the image. See e.g. https://buildkite.com/mrc-ide/orderly-dot-server-test/builds/10 or https://github.com/vimc/orderly.server/blob/master/buildkite/test-pipeline.yml 

It will push an image tagged with the sha but shouldn't affect `master` tag

hillalex:
> It will push an image tagged with the sha but shouldn't affect `master` tag

I wonder if we should re-think the tagging of those images? Seems a bit confusing that images with the same sha tag will actually be based off different versions of orderly... Like should we be tagging them with a tag that combines the `orderly` and `orderly.server` versions perhaps?

richfitz:
Agree with Alex here - this is something worth thinking about carefully, but I have no idea what the correct solution is...

r-ash:
Having the orderly.server test pipeline build the image with some combo of orderly and orderly.server versions sounds like a good idea. That pipeline is separate from the one run on orderly.server CI so having a different format for tagged images should mean we don't risk overwriting orderly server images. Will make a PR in orderly.server to tidy that up

richfitz (approved):

hillalex (approved):

r-ash:
Ok I think this is finally working as I want. You can see equivalent in this test repo (the actual build fails but due to permission issues, we can see the arg is setup correctly)


* On pull requests it triggers a build with the `orderly_version` set to the branch name see https://github.com/vimc/test-trigger/actions/runs/3395200404/jobs/5644929873#step:3:14
* On pushes to master it triggers a build with `orderly_version` set to `master` https://github.com/vimc/test-trigger/actions/runs/3395288038/jobs/5644973352#step:3:14

PR #325, opened 2023-01-30
Restore wd on bundle task failure

weshinsley:


weshinsley:
So, the setwd method is not ideal...

What's happening is that in orderly_version.R, around lines 599-605:-
```
 withr::with_envvar(private$envvar, {
        withr::with_dir(private$workdir, {
          source(private$recipe$script, local = private$envir, # nolint
                 echo = echo, max.deparse.length = Inf)
        })
      })
```

If the source call encounters a critical error, then control passes straight to the with CallingHandlers block around line 541-555, without resetting the working directory to what it was before the with_dir ->
```
  currentwd <- getwd()
      withCallingHandlers({
        ## Refetch the preflight info here: we want to keep git but
        ## replace everything else I think.  We might save the random
        ## seed but that is not actually supposed to work across R
        ## versions.
        private$preflight()
        private$preflight_info["git"] <- info$preflight_info["git"]

        self$run_execute(echo)
        self$run_cleanup()
      }, error = function(e) {
        setwd(currentwd)
        self$run_failed_cleanup(e)
      })
```

r-ash (approved):
Nice one

weshinsley:
Thanks Rob - let's merge this and hopefully we can then get more info on Paddy's cluster issue. @richfitz  - this ends up being a one-line fix which you did in the end, plus some tests - tagging you here as we mentioned it last week, re whether this is relevant for orderly3, or whether that migt be different enough anyway,.

